Issue #6, opened 2016-04-27
refresh keystore only when needed


panva:
At the moment a keystore refresh is called before every validation using that keystore, ideally validation is attempted, and only in case of a failure refresh is called. Refreshing should also be throttled.

PR #15, opened 2016-07-28
0.8.x 

panva:
- account is now configured similar to the adapter and has a "dev ready" default
- custom token endpoint grant registration
- encryption bugfixes


panva:
- resolves #6 
- resolves #15

PR #47, opened 2016-12-05
OAuth 2.0 for Native Apps BCP

panva:
A non-breaking feature flag implementing native client redirect uri validations to conform with three redirect_uri options defined in https://tools.ietf.org/html/draft-ietf-oauth-native-apps-06

Closes #46 

The way I interpret the spec, in regards to what the OP should handle.

I don't believe it's intended to be able to combine the three redirect URI options, i.e. custom uri scheme, with any hostname and dynamic ports. So, when the feature flag is provided and application_type is native the following presets should be possible:

* App-declared Custom URI Scheme Redirection
any non http/https uri scheme, with localhost as hostname and no dynamic ports

* App-claimed HTTPS URI Redirection
https with any valid (non loopback) hostname and no dynamic ports

* Loopback URI Redirection
http protocol with localhost, 127.0.0.1 or [::1] hostnames and dynamic port

And that boils down to:
- when custom scheme is given, localhost must still be used, no dynamic ports
- when https uri is given, any non loopback hostname can be used, no dynamic ports
- when http is used, any of the loopbacks must be used, and dynamic ports are allowed

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ The following drafts/experimental specifications are implemented by oidc-provide
 - [OpenID Connect Session Management 1.0 - draft 27][feature-session-management]
 - [OpenID Connect Back-Channel Logout 1.0 - draft 03][feature-backchannel-logout]
 - [RFC7592 - OAuth 2.0 Dynamic Client Registration Management Protocol (Update and Delete)][feature-registration-management]
-- [OAuth 2.0 for Native Apps - draft 06][feature-oauth-native-apps]
+- [OAuth 2.0 for Native Apps BCP - draft 06][feature-oauth-native-apps]
 
 Updates to drafts and experimental specification versions are released as MINOR library versions.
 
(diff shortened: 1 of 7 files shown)


ve7jtb:
As the person who wrote that part of dynamic client registration, I don't understand where you are getting the need to restrict custom scheme URI to having localhost  as the hostname.

The relevant sentence is:
Native Clients MUST only register  redirect_uris using custom URI schemes or URLs using the http: scheme with localhost as the hostname.

There is a or there so this may be a english parsing issue.
Native Clients MUST only register  (redirect_uris using custom URI schemes) or (URLs using the http: scheme with localhost as the hostname).

I will see if we can make that clearer in a errata.

John B.

panva:
Thank you John, i will correct the implementation mistake. Since you didn't mention any of the other options i'm going to go ahead and assume the rest of my interpretation and assumptions are correct.

Thank you.

PR #74, opened 2017-01-29
introspection & revocation authentication

panva:
- different client auth methods for introspection and revocation are now available
- fixed expected aud for for _jwt auth (introspection and revocation)


panva:
Not yet, in 2.0.

PR #157, opened 2017-10-03
removed ignore of expiration in token introspection

stekycz:


panva:
A test for covering the change in edge behavior would be great.

PR #160, opened 2017-10-10
Add explicit path to authorization/index.js to fix imports within Jest environments

wehriam:
Requiring `node-oidc-provider` in [Jest 20+](https://facebook.github.io/jest/) fails because [/lib/actions/authorization/index.js](https://github.com/panva/node-oidc-provider/blob/master/lib/actions/authorization/index.js) does not resolve correctly.

This change fixes the issue with the Jest 20+ resolver. See https://github.com/jsonnull/jest-directory-named-resolver for a custom Jest resolver that also addresses the issue.

All tests pass.


panva:
Could you please point me to the relevant jest issue?

wehriam:
I'm not aware of a Jest issue associated with this behavior. 

I'm also not sure if this is the expected behavior of the [default Jest resolver](https://github.com/facebook/jest/tree/master/packages/jest-resolve/src).

Possibly related:
* https://github.com/facebook/jest/issues/4567
* https://github.com/facebook/jest/pull/4315

panva:
my OCD for uniform code style can't stand this change :) How about what i just pushed, does that solve the jest issue for you?

Issue #164, opened 2017-10-26
update & consent scopes on subsequent requests


KtorZ:
Hi, I am trying to find a way to retrieve the scopes previously consented by a user upon a new authorization request. Here's a simple scenario explaining the issue:

1 - A non authenticated end-user is redirected to the authorization endpoint
2 - oidc-provider delegates the interaction to my server 
3 - The interaction completes with `{ login: { ... }, consent: { scope: 'openid my:scope' } }`
[ ... later ... ]
4 - The same end-user is redirected to the authorization endpoint with `scope=openid my:other:scope`

At that point, in a simple flow (no particular prompt, user already authenticated with an active and valid session in the AS), the `interactionCheck` function get called to see if any kind of interaction is needed.
In this case, I would like to reply that a `consent_required` interaction is needed since a new scope `my:other:scope` is requested and wasn't consented in the first request. 

However, I can't find a way to know from the oidc `Session` what scope was consented when the client was first authorized. Do I have to keep track of these authorized scopes for each client? To me, it sounds like something that `oidc-provider` should be doing but I might get the process wrong.

Thanks.

panva:
You're looking for the [`interactionCheck`](https://github.com/panva/node-oidc-provider/blob/master/lib/helpers/defaults.js#L348-L375) configuration option to specify this logic. You can return your consent_required again in that one. How you store and retrieve what has been consented to in the past is up to you, you may use the session object, custom cookie or even the your account object, whatever you have at your disposal.

KtorZ:
Yup, so I got the flow correctly. After the first authorization, the openid session looks roughly like this (using a MongoDB adapter):

```json
{
	"_id" : "93fe397b-1782-4534-b144-c64f789b67e0",
	"expiresAt" : ISODate("2018-10-26T21:17:10.415Z"),
	"account" : "479531f6-765a-4632-b5d0-6a47ece78227",
	"loginTs" : 1509029850,
	"authorizations" : {
		"myClient" : {
			"sid" : "e5852ab2-80a8-4f9b-9ccd-b1e0d539fded"
		}
	}
}
```

Wouldn't it make sense for the `oidc-provider` to also store consented scope next to the `sid` for each authorized client? Or do you consider this as a task my server should be managing (and therefore, keeps tracks of authorized clients and sessions myself which is actually part of the reason why I use `oidc-provider` in a first place).

panva:
[...]

PR #165, opened 2017-10-26
enable metadata passing in the 'result' object stored alongside the authorization sid

KtorZ:
See #164 

I went for a free object rather than storing directly the scopes as it is a bit more flexible. Let me know.

### DIFF
--- a/test/interaction/interaction.test.js
+++ b/test/interaction/interaction.test.js
@@ -325,9 +325,9 @@ describe('resume after interaction', () => {
 
       return this.agent.get('/auth/resume')
         .expect(() => {
-          const authorization = this.getSession().authorizations[config.client.client_id];
-          expect(authorization).to.have.property('meta');
-          expect(authorization.meta).to.have.property('scope');
+          const meta = this.getSession({ instantiate: true }).metaFor(config.client.client_id);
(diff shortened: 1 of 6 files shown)


test/interaction/interaction.test.js
```diff
@@ -305,6 +306,32 @@ describe('resume after interaction', () => {
...
+          remember: true,
+        },
+        meta: {
+          scope: 'openid',
+        },
+      });
+
+      return this.agent.get('/auth/resume')
+        .expect(() => {
+          const authorization = this.getSession().authorizations[config.client.client_id];
```
panva: can you use metaFor here? This would cover the last line for codecov

KtorZ: it seems that the `TestAdapter` doesn't return models' instances but raw objects; 

> this.getSession(...).metaFor is not a function

Any ideas ?

panva: no ideas, just test the metaFor function then.

KtorZ:
I've modified the `getSession` helper in the test suite to be able to get an instance of the Session object when necessary. This isn't performed by default since it would break a dozen of other tests that actually use the result of `getSession` as a plain JavaScript object.

Issue #167, opened 2017-10-27
primaturely end the interaction with an error


KtorZ:
Hi again,

On an other topic, I am trying to figure out a way to abort an interaction from within my authorization server with a specific error (for instance `access_denied` from the [OAuth 2.0 RFC](https://tools.ietf.org/html/rfc6749#section-4.1.2)). 

However, at the very best, it is possible to call the `interactionFinished` with an empty object, resulting in the `oidc-provider` redirecting the end-user towards the RP with an error `login_required` and description `End-User authentication is required`.  

I am missing something or is there currently no way to abort an interaction with a specific error code & description ?

panva:
This is not supported at the moment. Two ways you have, 

1. build a feature for this
2. you have the redirect uri available in your interactions, same as the response mode, you can handle the sendoff to the RP yourself

KtorZ:
My implementation currently relies on the second point indeed. However, while I redirect the user back to the RP, the interaction session remains active and open until expired. 

With the same spirit as #165, what would you think of enabling the AS to prematurely end the interaction with a singleton object `{ error: ... }`  containing an error code referenced in either of:

- http://openid.net/specs/openid-connect-core-1_0.html#AuthError
- https://tools.ietf.org/html/rfc6749#section-4.1.2.1

The description and state can be added by the `oidc-provider` as needed and it gives a clean way for the AS to handle errors (without having to manually build a url and perform the redirection).

panva:
I don't see how oidc-provider would add a description for an error it's not throwing, so let's provide an object with both error and error_description. Else I agree, state and valid response mode handling will be handled by the existing implementation.

KtorZ:
I don't see any room for custom error codes in the standard, which means that there's a fix list of values that could go in this `error` property and therefore, a corresponding list of descriptions that go with these errors. Don't you agree?

panva:
I don’t agree, someone will just return a custom error for his closed ecosystem provider and will want to provide a description of his own ;)

KtorZ:
I guess you're right. Let's go for an object: `{ error: ..., error_description: ... }` then to indicate a failed interaction. I will work on a PR.

panva:
PR Merged, to be released with 2.5.0 in the coming days.

KtorZ:
By the by, thanks for the diligence in handling those issues & PR :+1:

panva:
[...]

PR #168, opened 2017-10-27
Enable primature end of an interaction via an error result object

KtorZ:
See #167 

I've also added an optional  `error_status_code` field to the error object to be more consistent with other errors created by the provider. When not defined, it default to 400.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -265,6 +265,16 @@ router.post('/interaction/:grant', async (ctx, next) => {
 
   ['custom prompt name resolved']: {},
 }
+
+// optionally, interactions can be primaturely exited with a an error by providing a result
+// object as follow:
+{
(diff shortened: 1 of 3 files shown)


panva:
When is `error_status_code` ever useful? Authorization Errors will always result in a 302 back to the RP unless the problem is with clientid/redirect uri, this includes server_error afaik

KtorZ:
> When is error_status_code ever useful?

Well, I didn't dive much into what `ctx.throw(..)` does, so I thought of keeping the same error format while giving the ability to specify your own error code. I'll simply default to `400` all the time if it's actually not even used.

panva:
If i may, to keep it simple, don't allow the change of the code, always create an error with statusCode 400 and therefore expose true.

The responsible error handler is in lib/shared/authorization_error_handler.js

KtorZ:
done :+1:

PR #178, opened 2017-12-05
Allow system clock skew

davidgtonge:
This fixes an issue we had where client assertions and request objects
were being rejected because the systems clocks of the OP and RP
were out of alignment.

I used the same variable name and method of configuration as https://github.com/panva/node-openid-client


panva:
awesome contribution, thank you!

panva:
i will refactor to push the option to default configuration and document accordingly there.

PR #183, opened 2017-12-14
fix returning token introspection details for consumed token

stekycz:
The endpoint does not show any details about expired or revoked token. However, consumed token was returned with `active=false` but also with all details that could lead to security issues.

Possible BC break but no one should depend on it because it is not correct behavior.


panva:
brilliant @stekycz, thanks!

PR #188, opened 2018-01-04
Additional ID Token audiences helper

panva:
**audiences**

Helper used by the OP to push additional audiences to issued ID Tokens and other signed responses. The return value should either be falsy to omit adding additional audiences or an array of strings to push.  
affects: id token audiences, signed userinfo audiences  

default value:
```js
async audiences(ctx, accountId, token) {
  // token is a reference to the token used for which a given account is being loaded,
  // is undefined in scenarios where claims are returned from authorization endpoint
  return undefined;
}
```

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -687,6 +687,20 @@ default value:
 []
 ```
 
+### audiences
+
+Helper used by the OP to push additional audiences to issued ID Tokens and other signed responses. The return value should either be falsy to omit adding additional audiences or an array of strings to push.  
+affects: id token audiences, signed userinfo audiences  
+
+default value:
+```js
+async audiences(ctx, id, token) {
+  // token is a reference to the token used for which a given account is being loaded,
+  // is undefined in scenarios where claims are returned from authorization endpoint
+  return undefined;
+}
+```
+
 ### claims
 
 List of the Claim Names of the Claims that the OpenID Provider MAY be able to supply values for.  
@@ -841,7 +855,7 @@ default value:
 
 ### findById
 
-Helper used by the OP to load your account and retrieve it's avaialble claims. The return value should be a Promise and #claims() can return a Promise too  
+Helper used by the OP to load your account and retrieve it's available claims. The return value should be a Promise and #claims() can return a Promise too  
 affects: authorization, authorization_code and refresh_token grants, id token claims  
 
 default value:
--- a/lib/actions/authorization/process_response_types.js
+++ b/lib/actions/authorization/process_response_types.js
@@ -3,8 +3,8 @@ const instance = require('../../helpers/weak_cache');
 
 module.exports = (provider) => {
   const { IdToken, AccessToken, AuthorizationCode } = provider;
-  const pkce = instance(provider).configuration('features.pkce');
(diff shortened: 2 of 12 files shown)


panva:
Test Cov TBD
- Authorization Endpoint ID Tokens
- grant authorization_code ID Tokens
- grant refresh_token ID Tokens
- signed userinfo responses
- logout tokens

@MeenaAlfons feedback appreciated, please attempt to cover your use case.

panva:
@MeenaAlfons, any feedback?

MeenaAlfons:
The new feature works as expected. We've tried to use the new feature. The required audiences are added to the IdToken as expected. But our use case is not complete yet. Here is our problem.

The [`node-openid-client`](https://github.com/panva/node-openid-client) takes the client information in the constructor. When it tries to verify an IdToken, it not only checks for the signature but it also apply the **ID Token Validation** steps listed in the specs [3.1.3.7](http://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation).

- If the client is the same as in the azp claim, the validation passes successfully.
- If the client is different than the azp, the validation fails.

I understand that `node-openid-client` follows the specs but the specs has some confusing statements:

**[2. ID Token](http://openid.net/specs/openid-connect-core-1_0.html#IDToken)**
   `azp` OPTIONAL ... This Claim is only needed when the ID Token has a single audience value and that audience is different than the authorized party. It MAY be included even when the authorized party is the same as the sole audience.

**[3.1.3.7.  ID Token Validation](http://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation)**
   4. If the ID Token contains multiple audiences, the Client SHOULD verify that an azp Claim is present. 
   5. If an azp (authorized party) Claim is present, the Client SHOULD verify that its client_id is the Claim Value.

**Confusion:**
What does multiple `audiences` mean or what is it used for if the client (one of the audiences) will not be able to validate the ID Token unless the sole azp matches its client_id?

The multiple `audiences` seems useless for me. I understand that this question should go to the standard's guys but I hope you enlighten me.

panva:
> But our use case is not complete yet. Here is our problem.

I don't see it. azp is for the client to verify that the id token was issued for it in cases aud has more members. aud is for additional audiences which are not the clients. I think it's pretty self explanatory. You won't be able to use openid-client to validate id tokens issued for another client with yours as an audience. I suggest to use any jwt library for that where you can specify to validate the payload audience.

MeenaAlfons:
> additional audiences != clients, an id token is issued for one client, not more, if you have additional audiences included then they must not be seeking azp to be their id obviously.

This is the core point. I interpreted audiences == clients because they include `client_id`'s. But from your description it seems that the distinction between them is real and _extra audiences_ are not treated as _clients_ in this context.

panva:
No they are indeed not. Audiences are not limited to client_ids that the OP recognizes. You push your Resource Server identifiers in there, additional client ids maybe too, but it is not limited to those. I hope it's clarified now. There's no mechanism that would make an ID Token validate for two openid clients including the `azp` claim. If you wish to accept them for other entities just have them inspect the aud claim, not azp.

PR #191, opened 2018-01-17
Added new configuration property to allow unsafe redirect uris

kay-schecker:
In our project we use your oidc provider as internal test server. For that it would be very useful to allow unsafe redirect_uris, because we want to use localhost and http for local development.
Please see the updated configuration.md for more details.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -106,6 +106,29 @@ const clients = [
 oidc.initialize({ clients }).then(fulfillmentHandler, rejectionHandler);
 ```
 
+If you use the "implicit" `grant_type` and the `application_type` "web" (default), then `redirect_uris` with the scheme 
+*http* or the host *localhost* are rejected by default. To enable unsafe `redirect_uris` you can set the configuration
+property `allow_unsafe_redirect_uris` to true. **Do not activate this behavior in production!**
+
+```js
+const oidc = new Provider('http://localhost:3000');
+const clients = [
+  {
+    allow_unsafe_redirect_uris: true, // Do not activate this behavior in production!
+    token_endpoint_auth_method: 'none',
+    client_id: 'mywebsite',
+    grant_types: ['implicit'],
+    response_types: ['id_token'],
+    redirect_uris: ['http://localhost/cb'],
+  },
+  {
+    // ...
+  },
+];
+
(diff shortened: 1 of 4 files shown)


kay-schecker:
One of the three builds has failed, but it seems not to be related to my changes?

panva:
Hello @kay-schecker, while I understand the use case and already received these kinds of requests in the past, it is not a feature i wish to offer, not with an env variable, neither with a configuration setting and for sure not as part of recognized client metadata. I'm afraid you'll have to work off a fork or find a another solution. Thank you for understanding, ignoring the specs and allowing this for dev environment will only lead to devs forgetting these options are in place and end up with vulnerable client setups

kay-schecker:
Hi @panva, thanks for your fast reply. I clearly understand your reasoning, but isn't there any other way besides the fork, because this will potentially end in a nightmare of forks, if other developers need the same feature. It would be great to have plugin support for the oidc-provider, so developers can add functionality like this easily.


Regarding your suggestion to use TLS certificates on our dev-servers:

We work with [angular-cli](https://github.com/angular/angular-cli) and start a server on the developer machine itself. This server isn't protected by TLS by default. Furthermore the application is served on 0.0.0.0 / localhost then. That means that we'll run always into the same issue, when we start a new [angular-cli](https://github.com/angular/angular-cli) project. Unfortunately, Angular also has a [bug ](https://github.com/angular/angular-cli/issues/5826) at the moment, so we can't start with TLS at all. But i know this doesn't matter for the oidc-provider. I'm just looking for a solution to my problem.

panva:
I really don't want a feature like this in the code and documentation. I can see the following, and let me know if it works for you.

The `Client` constructor that's returned uses a provider specific `Schema`, this Schema may be assigned to the Client as a class property. Through that you could simply overload the `redirectUris()` schema method to not do any validations. This is hopefully destructive enough that noone will actually use aside of test environments.

kay-schecker:
Thanks a lot i'll give it a try and let you know :-)

kay-schecker:
Yeah, it's working well 👍 Do you plan a new release or a snapshot to publish this change?

panva:
Probably next week, until then just use master.

Issue #228, opened 2018-03-15
TypeScript


ulrikstrid:
Any thoughts on TypeScript?

I'm using this lib right and as I'm working I'm building some types for it. Is that something you would be interested in taking as a contribution?

I would also be willing to contribute to a rewrite in TS if your interested.

panva:
Hi @ulrikstrid,

unfortunately I do not have the free bandwidth right now to do a rewrite, plus I don't see the point.

As I'm not familiar with the TypeScript ecosystem, could you elaborate on what maintaining types would entail? Is having types exposed the bare minimum for TypeScript projects to be able to work with oidc-provider?

ulrikstrid:
Maintaining external types might be painful if you want to change the API a lot but I guess that will mostly be between versions. I have the following types right now that at least lets me get going with TypeScript but is no where near complete.

```ts
declare module "oidc-provider" {
  import { Request, Response, Handler } from "express";

  class Provider {
    app: {
      proxy: boolean;
    };

    uuid: string;

    constructor(url: string, {});

    initialize(config: {}): Promise<void>;

    interactionDetails(
      req: Request
    ): Promise<{
      interaction: {
        reason: "consent_prompt" | "client_not_authorized";
      };
    }>;

    interactionFinished(req: Request, res: Response, {}): Promise<void>;

    callback: Handler;
  }

  export = Provider;
}
```

I have some time to spend, if you want I could just change a few files and open a PR to show the difference.

panva:
You can give it a shot, but I cannot promise to take it over and maintain it given i'm not a TypeScript user myself.

ulrikstrid:
Of course, that's why I want to start with just a few files. 👍

I will try to get this done in during the day so that you can give a thumbs up or down.

panva:
Looking at the PR this seems totally not worth a rewrite :)

ulrikstrid:
I agree, I would rather maintaining typings for the parts that I use. But it was worth a shot

panva:
no worries

MichaelCastiau:
I would actually really appreciate Typescript support for this library. Working in a production environment with virtually every node app in written in Typescript, it is a pain to have to move back to a library that only supports plain JavaScript (aka no intellisense support). 

[...]

PR #230, opened 2018-03-16
Typescript

ulrikstrid:
This is very much a work in progress and it seems like it will take a lot of work to get something working properly. The tests doesn't run as it is now.
I started with the models folder just to start somewhere.
I have committed the compiled .js files as well just to see the difference .


panva:
As discussed in #228

Issue #234, opened 2018-03-25
Make PKCE supported challenge methods configurable


psmiraglia:
According to the current code, PKCE enabling is more or less a boolean setting. If true, the OP enables the support for `plain` and `S256` challenge methods. Unfortunately, this approach makes the OP unable to operate in an iGov context where `plain` is not allowed (see https://xml2rfc.tools.ietf.org/cgi-bin/xml2rfc.cgi?Submit=Submit&format=ascii&mode=html&type=ascii&url=https://bitbucket.org/openid/igov/raw/HEAD/openid-igov-oauth2.xml#rfc.section.3.1.7). It would be nice to have this setting like that:

```javascript
// enables PKCE with support for both "plain" and "S256" (already available)
const configuration = { features: { pkce: true } };
```

```javascript
// enables and force the use of PKCE with support for both "plain" and "S256" (already available)
const configuration = { features: { pkce: { forcedForNative: true } } };
```

```javascript
// enables the use of PKCE with support for just "S256" (desired feature)
const configuration = { features: { pkce: ['S256'] } };
```

What do you think about that? From the dev perspective, I think major changes should occur under `node-oidc-provider/lib/actions/authorization/check_pixy.js`.

psmiraglia:
With this new feature I think `pkce` could be always managed as `Object`

```javascript
const configuration = {
  features: {
    pkce: {
      supportedMethods: ['plain', 'S256'],
      forcedForNative: true,
    }
  }
}
```

panva:
I like the feature request. Wasn't aware of the iGov profile requirements. Will do in the next couple days. 

That being said you can already get it going that way today.
Post middleware to remove the plain value from discovery and a pre middleware to throw when plain is used.

psmiraglia:
Nice! If interested, I'm already drafting the solution...

https://github.com/psmiraglia/node-oidc-provider/tree/issue/234

panva:
Looked briefly, you shouldn't assume the default is S256 if nothing is provided. iGov only says that plain must not be used, the way a method is passed in remains unchanged and so S256 as a method should be provided by clients explicitly.

panva:
Couple of changes i made:
- Default values for features are to be booleans, on/off toggles, when on and in need of fine-tuning then they may be objects.
- Switched the order of check / default code_challenge_method to stick to default plain but also allow not supporting it with an appropriate error.

It'd be great if you could do some testing.

PR #236, opened 2018-03-26
Add some tests

psmiraglia:
Contribute for #234


lib/actions/grants/authorization_code.js
```diff
@@ -25,6 +25,10 @@ module.exports.handler = function getAuthorizationCodeHandler(provider) {
 
     // PKCE check
     if (pkce && (ctx.oidc.params.code_verifier || code.codeChallenge)) {
+      if (!pkce.supportedMethods.includes(code.codeChallengeMethod)) {
+        ctx.throw(new InvalidGrantError('challenge method is not supported'));
+      }
+
```
panva: In what case would this codepath be executed? Code is only pushed to the DB by regular path (non-tests) after the exact same validation, having it again here is redundant.

Issue #238, opened 2018-03-27
Fine tuning on recognized client metadata


psmiraglia:
According to [OpenID Connect Registration specs](https://openid.net/specs/openid-connect-registration-1_0.html#ClientMetadata)

> [...] Additional Client Metadata parameters MAY also be used. Some are defined by other specifications, such as OpenID Connect Session Management 1.0 [...]

Does the current implementation permit a fine-tuning on recognized client metadata parameters According to the following code, the list of recognized parameters is not tunable.

https://github.com/panva/node-oidc-provider/blob/146136acac81c027182706512b6004fa5af8131d/lib/consts/client_attributes.js#L1-L28

Why not having something like that?

```js

provider.initialize({
  extraClientMetadata: [
    'my_extra_attr_1',
    'my_extra_attr_2',
  ],
})
```

panva:
> Does the current implementation permit a fine-tuning on recognized client metadata parameters?

It does not by any public API.

NB: I'm pretty sure you can just require the client_attributes upfront, add attributes and go. Has the same value as what you're proposing.

psmiraglia:
>  I'm pretty sure you can just require the client_attributes upfront, add attributes and go. Has the same value as what you're proposing

Could you provide me with an example?

panva:
> Could you provide me with an example?

Simple actually:

```js
const client_attributes = require('./node_modules/node-oidc-provider/lib/consts/client_attributes.js');
client_attributes.RECOGNIZED_METADATA.push('foo');
client_attributes.REQUIRED.push('foo');
client_attributes.BOOL.push('foo');

const Provider = require('oidc-provider');
```

Either this or further work on #239 to make it possible scoped for each instance.

psmiraglia:
Oh sure...

About #239, what do you think about that?

```js
provider.initialize({
  extraClientMetadata: {
    extra_attribute: function itsOwnValidator(value) {
      let isValid = flase;      
      // do something...
      return isValid;
    }
  }
})
```

panva:
That's a good start, but passing in just value won't do it tbh - i'd say you also need all the other metadata and somehow have the option to set values for other metadata from these functions.

It should also be a part of the `new Provider` constructor call, not `initialize` and should be called for each property that's registered, not just the ones with values.

```
new Provider('...', {
  extraClientMetadata: {
    properties: ['foo', 'bar'],
    async validator(key, value, metadata) { // will be called for foo and bar regardless of the provided values.
      // validations
[...]

PR #239, opened 2018-03-27
Fine tuning on recognized client metadata

psmiraglia:
Contribute to #238

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -917,6 +917,14 @@ default value:
 }
 ```
 
+### extraClientMetadata
+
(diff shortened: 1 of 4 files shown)


panva:
These properties to me seem to have no validations, no processing, no use. If the implementation won't use them, what's the point of having them?

psmiraglia:
Your comments describe a significant change (the `Provider` constructor). I started playing with your "toy" just two weeks ago and, to be honest, I don't know if my knowledge about the module is enough to produce good and safe code in few time. Anyway, I'll try to write something during Easter holidays.

panva:
I'll take a stab at it then.

PR #240, opened 2018-03-28
extraClientMetadata config option

### LINKED ISSUE #238: Fine tuning on recognized client metadata

panva:
Closes #238

### DIFF
--- a/lib/helpers/client_schema.js
+++ b/lib/helpers/client_schema.js
@@ -146,7 +146,11 @@ module.exports = function getSchema(provider) {
         }
       });
 
-      Object.assign(this, DEFAULT, _.pick(metadata, RECOGNIZED_METADATA));
+      Object.assign(
+        this,
+        DEFAULT,
+        _.pick(metadata, ...RECOGNIZED_METADATA, ...configuration.extraClientMetadata.properties),
+      );
 
       this.required();
       this.whens();
@@ -235,6 +239,9 @@ module.exports = function getSchema(provider) {
           invalidate('jwks.keys must not be empty');
         }
       }
+
+      this.processCustomMetadata();
+      this.ensureStripUnrecognized();
     }
 
     required() {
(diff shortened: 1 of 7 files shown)


lib/helpers/client_schema.js
```diff
@@ -427,6 +434,22 @@ module.exports = function getSchema(provider) {
...
+    processCustomMetadata() {
+      configuration.extraClientMetadata.properties.forEach((prop) => {
+        configuration.extraClientMetadata.validator(prop, this[prop], this);
+      });
+    }
+
+    ensureStripUnrecognized() {
+      const whitelisted = [...RECOGNIZED_METADATA, ...configuration.extraClientMetadata.properties];
+      Object.keys(this).forEach((prop) => {
+        if (prop === 'sector_identifier') return;
```
psmiraglia: Why `sector_identifier` is a special case?

test/configuration/custom_client_metadata.test.js
```diff
@@ -0,0 +1,153 @@
...
+          redirect_uris: ['http://rp.example.com/cb'],
+        },
+      ],
+    });
+
+    const client = await provider.Client.find('client');
+    expect(client).not.to.have.property('foo');
+  });
+
+  it('can be used to add validations to existing standard properties', async () => {
```
psmiraglia: Very interesting case.

lib/helpers/client_schema.js
panva: `sector_identifier` is an internal property, not one passed by clients - it is resolved instead.

PR #253, opened 2018-04-14
Colorize console output requiring attention

madarche:
It only colors the output if in a TTY

The technique used comes from https://github.com/dougwilson/nodejs-depd

### DIFF
--- a/lib/helpers/defaults.js
+++ b/lib/helpers/defaults.js
@@ -1,4 +1,5 @@
 const LRU = require('lru-cache');
+const attention = require('../helpers/attention');
 const epochTime = require('../helpers/epoch_time');
 
 const cache = new LRU(100);
@@ -7,7 +8,7 @@ const warned = new Set();
 function changeme(name, msg) {
   if (!warned.has(name)) {
     warned.add(name);
-    console.info(`NOTICE: default helper ${name} called, you should probably change it in order to ${msg}.`); // eslint-disable-line no-console
+    attention.info(`default helper ${name} called, you should probably change it in order to ${msg}.`);
   }
 }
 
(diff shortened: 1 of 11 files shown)


madarche:
@panva: OK. This will be ready in a few days. Don't close :-)

lib/helpers/defaults.js
```diff
@@ -3,11 +3,19 @@ const epochTime = require('../helpers/epoch_time');
...
+
 const warned = new Set();
 function changeme(name, msg) {
   if (!warned.has(name)) {
     warned.add(name);
-    console.info(`NOTICE: default helper ${name} called, you should probably change it in order to ${msg}.`); // eslint-disable-line no-console
+
+    let str = `NOTICE: default helper ${name} called, you should probably change it in order to ${msg}.`;
+    if (process.stdout.isTTY) {
+      str = SET_BOLD_YELLOW_TEXT + str + RESET_ALL_ATTRIBUTES;
```
panva: `${SET_BOLD_YELLOW_TEXT}${str}${RESET_ALL_ATTRIBUTES}` instead

madarche:
I'll improve the code coverage.

madarche:
@panva: I was tempted to direct all *attention* logs to `console.warn`, because that would make all those logs go down to `stderr`, and in my opinion the *changeme* logs should go to `stderr` too. Having all those logs in `stderr` is nice with process management tools like [PM2](http://pm2.keymetrics.io/docs/usage/log-management/). What do you think?

panva:
changeme is not a directive, a must, some of them are purely informational and they need not be changed. Using dev keys on the other hand, is a different level log information.

PR #257, opened 2018-04-22
v4.x

panva:
Improvements
- [x] improve audiences helper use cases (more params, also for access token)
- [ ] add LRU cache sizes config
- [ ] session store changes (exp)
- [ ] idle session timeout
- [ ] unified data model (maybe, finally, no idea how yet)
- [ ] oauth2 mode (no openid scope) - maybe
- [x] no default "session" acr
- [x] ^ add a note on how to backfill the existing behavior
- [x] thrown errors should be instances of what is exposed instead of `http-errors`

Fixes
- [x] refactor and fix issues with `find_client_id.js` and following middlewares
- [x] fix client jwks_uri refresh error to be invalid_client_metadata and propagated to the client
- [x] don't ignore `null` values
- [ ] ^ add a note on how to migrate existing stored clients
- [x] wrong `token_type` values from introspection, remove the claim
- [x] ^ add a note on how to backfill the existing values

Configuration
- [ ] all configuration read on runtime when needed
- [ ] get/set/enable/disable/initialize configuration

Maintenance
- [x] remove deprecations


slavab89:
Looking forward for the `improve audiences helper use cases` 😄 
Do you think it would be possible to include audience in client_credentials grant as well? is it a valid property to have in that flow?

panva:
@slavab89, i’ll give it a shot. The accountId will be undefined tho. 

Anything else you’d like to see, I might consider it, ...

panva:
#269

Issue #258, opened 2018-04-24
Extra parameters are ignored in the client auth flow


nataliahering:
The [extra parameters](https://github.com/panva/node-oidc-provider/blob/master/docs/configuration.md#extraparams]) are available in the `ctx.oidc.params`, only in the implicit flow. 
It has been observed that they are not being added in the 2-leg auth flow. 

The whitelist that is being created [here](https://github.com/panva/node-oidc-provider/blob/master/lib/shared/chains/client_auth.js) does not include the extra params from the configuration and hence they are getting ignored.

We have the need to pass extra information in the 2-leg flow. Possible solutions:
1) During the first step, store the extra params value in the DB (sessions)
2) Update the `client_auth` and whitelist also the extra params 
`const extras = instance(provider).configuration('extraParams');`
 `extras.forEach(whitelist.add.bind(whitelist));`
  `const buildParams = params(whitelist);`
3) Use the client_secret_jwt auth and pass the extra information as extra claims in the jwt token. This can be decoded and the information can be added to the `ctx.oidc.params`.

The (2) and (3) have the drawback that the extra information can't be passed from the first step.

Any ideas?

panva:
> The extra parameters are available in the ctx.oidc.params, _only in the implicit flow_.

That's incorrect, they're being parsed and available for interactions and helpers such as `interactionCheck` in every `authorization_endpoint` request regardless of the response_type. The original intention was so that e.g. tracking parameters can be passed to interactions for trusted clients and then displayed in the views.

It is working exactly as documented.

```
extraParams
Pass an iterable object (i.e. Array or set of strings) to extend the parameters recognized by the **authorization endpoint**. These parameters are then available in ctx.oidc.params as well as passed to interaction session details

affects: **authorization, interaction**
```

I'm not sure what you're trying to achieve, which grant_type are trying to change/are implementing. Please provide more details.

nataliahering:
Thanks for the prompt reply. Here are some more details.
Assume we have set up the config to include `extraParams: ['extraParameter']`.
Flow:
step 1: `GET` to `oidc/auth` endpoint, `response_type=code`, `extraParameter=extraValue`
step 2: `POST` to `oidc/token` endpoint, using the `client_secret_basic` flow and the code returned from step 1. Include the extraParameter=extraValue in the body.

[...]

PR #259, opened 2018-04-25
Add extraParams in the client_auth whitelist

nataliahering:


panva:
See #258, extraParams is only intended (and documented as such) to enrich authorization and interaction contexts.

PR #266, opened 2018-05-01
Fix miss of grant object destroy

### LINKED ISSUE #264: Shouldn't the grant object be destroyed when its grant related models are destroyed?
If I'm not mistaken, currently in the proposed [in-memory adapter](/lib/adapters/memory_adapter.js),  [redis adapter](/example/adapters/redis.js) and [mongodb adapter](/example/adapters/mongodb.js) the grant object (holding the references to the grant related models) is not destroyed when a model is destroyed.

Shouldn't this grant object be destroyed as well?

This is not a bug in the oidc-provider lib, but since developers are expected to use those proposed adapters as reference this can be misleading.

madarche:
This fixes #264.

### DIFF
--- a/lib/adapters/memory_adapter.js
+++ b/lib/adapters/memory_adapter.js
@@ -24,8 +24,8 @@ class MemoryAdapter {
 
     if (grantId) {
       const grantKey = grantKeyFor(grantId);
-
       storage.get(grantKey).forEach(token => storage.del(token));
+      storage.del(grantKey);
     }
 
     return Promise.resolve();
--- a/test/configuration/adapter_test.test.js
+++ b/test/configuration/adapter_test.test.js
@@ -4,22 +4,19 @@ const { TestAdapter } = require('../models');
 const { AdapterTest } = Provider;
 
 describe('AdapterTest', () => {
-  it('passes with the default MemoryAdapter', () => {
+  it('passes with the default MemoryAdapter', async () => {
     const provider = new Provider('http://localhost');
-
-    return provider.initialize().then(() => {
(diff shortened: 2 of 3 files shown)


panva:
Can you avoid the introduction of the class getter? I feel this is unnecessary. How you keep track of the grant chain and cleaning it up is every implementer’s choice.

madarche:
Having the low-level storage class getter is useful for testing and this is really the main point of it. But this might also be useful for doing more specific stuffs that the default adapter interface doesn't offer, like getting all the records of one model.

The only downside of it I can think of is that the  low-level storage class getter gives an open access to the database used by the adapter. This should not be a security issue but it might be better to avoid it.

So @panva do you confirm you want the class getter avoided and thus no test on the grant object destroy?

panva:
Definitely. It is an unnecessary getter the library has no use for. Take mongo, there’s no concept of a grant object. A redis one can also be modeled differently to avoid this. And the tests i have better plans for anyway.

madarche:
OK @panva, understood. I'll do it in the next couple of hours.

lib/adapters/memory_adapter.js
```diff
@@ -24,8 +24,8 @@ class MemoryAdapter {
 
     if (grantId) {
       const grantKey = grantKeyFor(grantId);
-
       storage.get(grantKey).forEach(token => storage.del(token));
+      storage.del(grantKey);
```
panva: I think all that's needed is this and a similar line in redis_adapter example.

madarche:
@panva, done, I have removed all the code regarding the class getter.

PR #269, opened 2018-05-02
v4.x

panva:
Improvements
- [x] improve audiences helper use cases (more params, also for access token)
- [x] session store changes (exp)
- [x] no default "session" acr
- [x] ^ add a note on how to backfill the existing behavior
- [x] thrown errors should be instances of what is exposed instead of `http-errors`
- [x] upgrade node-jose to ^1.0.0

Fixes
- [x] refactor and fix issues with `find_client_id.js` and following middlewares
- [x] don't ignore `null` values
- [x] ^ add a note on how to migrate existing stored clients
- [x] wrong `token_type` values from introspection, remove the claim
- [x] ^ add a note on how to backfill the existing values

Maintenance
- [x] remove deprecations


madarche:
Hello @panva. Since you are working on the errors, would it also be possible that:
* the `renderError` method now returns a true error instead of an object?
* those errors include primarily a `code` (along with the message) like in Node.js so they are easier to  check cf. https://medium.com/the-node-js-collection/node-js-errors-changes-you-need-to-know-about-dc8c82417f65

At the moment I have the code below and I doesn't feel good:
```javascript
async renderError(ctx, error) {
  if (error.error == 'invalid_request' && error.error_description == 'could not find logout details') {
    // Deal with it
  }
}

Thanks!

**Comment by SokratisVidros** [comment]
👍 for the suggestion both for error_handler and authorization_error_handler. Currently we also use workarounds such as:

```js
function handleHttpError(ctx, error) {
  if (!(error instanceof Error)) {
    error = new Boom(`${error.error} - ${error.error_description}`, {
      statusCode: ctx.status
    });
  }
  ...
}
```

panva:
The method is not for handling but purely rendering for an end-user, should the environment be a browser one. I doubt server errors are in any way communicateable to users, for logging them you can use `provider.on('server_error', ...))`, which you probably have to have registered anyway for the cases this is an API call.

Anyway, @madarche parsing error descriptions, bad idea - these can always change. Instead why don't you check the go off ctx._matchedRouteName, ctx.method and invalid_request only?

madarche:
@panva 

I know that parsing error description is a bad idea. That's the point of my request here: to be able to have an efficient and safe way to react differently upon errors.

Thanks for the suggestion concerning the check to do, but the following, if safer (thanks) …

```javascript
async renderError(ctx, error) {
  if (ctx._matchedRouteName == 'end_session' && ctx.method == 'POST' && error.error == 'invalid_request' ) {
    // Explain to the user
  }
}
```

… doesn't feel as efficient as this code would:

```javascript
async renderError(ctx, error) {
  if (error.code == 'session_not_found') {
    // Explain to the user
  }
}
```

I understand that the `renderError` method is only, well, to render an error, that is only to explain to the user.

But what's more efficient to convey this information to a UI layer, be it a browser, than an `Error` object with a unique code?

Moreover one expects an `error` argument in a method to be an instance of `Error`.

I'm not trying to be troublesome, but since v4 has some breaking changes and some changes in this area, this is really a good time to try to improve this, isn't it?

panva:
I will add an argument that's the actual error object and adjust the docs. The second arg is just the `out` object anyway and will remain as such. More than this isn't necessary for rendering. But i see the point and this will be the new signature.

```
async renderError(ctx, out, error) {
```

This however won't give you a unique code. If you're up for this, once 4.x is out you can give a shot to a PR that introduces unique codes for every error instance in the library.

That would be a great step towards providing error_uri too.

madarche:
Perfect @panva! And of course I'm not only complaining and asking, I'll work on this PR for unique error codes after 4.x is out! Cool!

SokratisVidros:
This solution caters for most needs. Thanks @panva.

PR #271, opened 2018-05-07
In bookshelf postgresql adapter, fix expired records purge and grant object removal

madarche:
Now all expired records are purged, not only those which are asked for through a find.

And the grant object is now removed.


madarche:
Thanks @panva!

Issue #221, opened 2018-03-05
/session/check relies on availability of github


nico-schunke:
Hi Filip,

we just noticed that the helper to check whether it is allowed to set thirdparty cookies (https://github.com/mindmup/3rdpartycookiecheck) ist hard coded at the /session/check action: https://github.com/panva/node-oidc-provider/blob/master/lib/actions/check_session.js#L86

To be honest I do not want to overwrite the whole logic to replace that url, called by the IFrame. Do you think there is an easy way to configure?

(btw: same applies for the sha256 script: https://github.com/panva/node-oidc-provider/blob/master/lib/actions/check_session.js#L23)

Personally I would like to host that scripts on my own CDN and take care about the availability by my own.

Thanks for feedbacking!

Best Nico

panva:
If you wish to host these on your own (knowing the implications, such as the point of the 3rd party cookie check actually being on a different domain for it to be able to do what it does) you can [register a middleware](https://github.com/panva/node-oidc-provider/blob/master/docs/configuration.md#pre--and-post-middlewares) in the koa app stack and change the response body. Such as

```
provider.app.middleware.unshift(async (ctx, next) => {
  await next();

  if (ctx._matchedRouteName === 'check_session') {
    ctx.body = ctx.body.replace(...) // or similar, modifying the body replacing the script URIs
  }
});
```

Currently I don't plan to offer option to build this in unless there's a strong demand for it.

nico-schunke:
Thanks for your answer. Your suggestion would work for me, tough it seems like to be a bit "hacky" to me. But anyways it does it fit my needs.

panva:
You might say it is, but there's unfortunately not enough bandwith for me to give a configuration option for everything. If you want you could submit a PR where these external resources are in the configuration and can be overwritten. Your call.

nico-schunke:
Understood. We'll think about when we come across again. Right now it is not really a showstopper. Thanks for your efforts and taking care.

PR #273, opened 2018-05-07
Make third party check helper url configurable

stavros-wb:
I've added a configuration option for linking the 1st step of third party cookies support helper. The github dependency is ok for development and testing, but I think every user using this lib in production should provide their own hosting (making sure the host isn't the same as the OIDC one).

https://github.com/panva/node-oidc-provider/issues/221#issuecomment-370734677

> (As a final note — don’t use someone else’s server to test third-party cookies without their permission. It could break spontaneously or inject malware. And it’s rude.)
https://stackoverflow.com/a/7104048

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -518,9 +518,10 @@ To disable removing frame-ancestors from Content-Security-Policy and X-Frame-Opt
 const configuration = { features: { sessionManagement: { keepHeaders: true } } };
 ```
 
-In order for the Session Managmenet feature to work, the User-Agent must allow access to Third-Party 
-cookies. oidc-provider checks if this feature is enabled using a [CDN hosted](https://rawgit.com/)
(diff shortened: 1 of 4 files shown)


panva:
Thank you,

I think that 1) this could use way more documentation (setup, background) if you expect others to host it. 2) isn’t that big of an issue given that the pages are hosted off a cdn and essentially come from the same source the library itself does 3) it, if you think it really important, should log a warning when using the default.

stavros-wb:
@panva can you please take a look at the changes?

panva:
I can, in the coming days.

stavros-wb:
Thank you @panva

panva:
Can you close, rebase and reopen this for v4.x-wip?

stavros-wb:
sure

PR #294, opened 2018-06-16
storage and token formats support (legacy, opaque, JWT)

panva:
Added `formats` configuration option. This option allows to configure the token storage and value formats. The different values change how a token value is generated as well as what properties get sent to the adapter for storage. Three formats are defined (see [`lib/models/formats`](https://github.com/panva/node-oidc-provider/tree/storage-formats/lib/models/formats))

- `legacy` is the current and default format until next major release. no changes in the format sent to adapter
- `opaque` formatted tokens have a different value then `legacy` and in addition store what was in
  legacy format encoded under `payload` as root properties, this makes analysing the data in your storage way easier
- `jwt` formatted tokens are issued as JWTs and stored the same as `opaque` only with additional property `jwt`. The signing algorithm for these tokens uses the client's `id_token_signed_response_alg` value and falls back to `RS256` for tokens with no relation to a client or when the client's alg is `none`

This feature uses the previously defined public token API of `[klass].prototype.getValueAndPayload, [klass].prototype.constructor.getTokenId, [klass].prototype.constructor.verify` and adds a new one `[klass].prototype.constructor.generateTokenId`. See the inline comment docs for more detail on those.

Example use, setting default opaque and access tokens as JWTs.
```js
const oidc = new Provider('...', {
  formats: { default: 'opaque', AccessToken: 'jwt' },
});
```

**Further format ideas and suggestions are welcome.**

@madarche please check and adapt `example/adapters/bookshelf_postgresql_adapter.js` for new storage formats support (opaque and jwt) if needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ Yay for [SemVer](http://semver.org/).
 
 **Table of Contents**
 
-<!-- TOC START min:2 max:2 link:true update:true -->
+<!-- TOC depthFrom:2 depthTo:2 withLinks:1 updateOnSave:1 orderedList:0 -->
+
 - [Unreleased](#unreleased)
 - [4.0.x](#40x)
 - [3.0.x](#30x)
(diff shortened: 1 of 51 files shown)


madarche:
@panva I'll look shortly (max a couple of days). Other projects are eating all my time at the moment :confused:

Issue #315, opened 2018-07-16
Inconsistent mountPath


cymarechal:
provider.js (line 120):
`instance(this).mountPath = url.parse(this.issuer).pathname;`

`http://localhost/OIDC` -> `/OIDC`
`http://localhost/OIDC/` -> `/OIDC/`

provider.js (line 196):
`return [mountPath !== '/' ? mountPath : undefined, routerUrl].join('');`

May return `/OIDC//auth/:grant:`, which is not a valid URL.

panva:
Don't mount 🤣, no really, thanks for the report @cymarechal, would you be able to write a failing test and submit a fix? Else, i'm happy to look into it when there's time.

cymarechal:
I'm on it! 👍

PR #316, opened 2018-07-16
fix: mountPath

cymarechal:
Solves #315 .

-> Need review on lib/provider.js:89. (feels dirty?)


panva:
Let’s just fix the path/url generation without modifying the issuer string.

cymarechal:
@panva Better to remove "/" at the end of mountPath variable (pathFor - provider.js:188) or to replace "//" by "/" in the return statement? Or we could even edit mountPath globally?

panva:
updated a bit, didn't realize your fork is master but i force-pushed to it, can you review?

Issue #350, opened 2018-08-13
interactionFinished custom response handling


JSantha:
I would like to implement login interaction in AJAX way and handle redirection on client side. This is not easy to achieve since `provider.interactionFinished` has standard 302 redirect hardcoded.

```js
async interactionFinished(req, res, result) {
    const interaction = await getInteraction.call(this, req, res);
    interaction.result = result;
    await interaction.save(interaction.exp - epochTime());

    res.statusCode = 302; // eslint-disable-line no-param-reassign
    res.setHeader('Location', interaction.returnTo);
    res.setHeader('Content-Length', '0');
    res.end();
}
```
 
It's off course possible to monkey-patch response to achieve AJAX behaviour but it's not obviously a proper solution:

```js
const origEnd = res.end;
res.end = function (this: Response): Response {
    res.end = origEnd;
    const location = this.getHeader('Location');
    this.removeHeader('Location');
    this.removeHeader('Content-Length');
    this.statusCode = 200;

    return res.send({ location });
};
```

Does OIDC allow custom response handling on interaction finished? Or delegating redirection responsibility to client side can imply new security issues?

panva:
This has nothing to do with OIDC, it's internal handling. The function is merely storing what you pass in the interaction object's results and then redirects to the returnTo path. You can easily read what the function does and not use it at all, do it on your own. The only important thing is that after your ajax call finishes you redirect to that location.

JSantha:
Thank you, I thought about this but got afraid of properly setting expiration on save since I need to "reimplement" calculation of expiration `interaction.exp - epochTime()` in my application code.

Maybe it would be great to extract following lines to independent method:

```js
const interaction = await getInteraction.call(this, req, res);
interaction.result = result;
await interaction.save(interaction.exp - epochTime());
```
So I can reuse this logic but handle response by myself. What do you think?

panva:
Sure.

JSantha:
Created PR #351.

PR #351, opened 2018-08-13
feat: option to setInteractionResult without calling interactionFinished

JSantha:
As already mentioned here #350. We have no clear way to propagate interaction result without causing response redirect. To allow custom response handling I'm submitting this non-breaking PR which simply extracts result propagation to its own method.

Now, instead of calling `provider. interactionFinished` we have option to call `provider.setInteractionResult` with same arguments to prevent immediate redirection.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -320,6 +320,27 @@ router.post('/interaction/:grant', async (ctx, next) => {
 }
 ```
 
+**`#provider.interactionResult`**
+Unlike `#provider.interactionFinished` redirect uri is returned instead of immediate http redirect.
+It should be used when custom response handling is needed e.g. making AJAX login where redirect uri 
+is expected to be available in the response.
+
+```js
+// with express
+expressApp.post('/interaction/:grant/login', async (req, res) => {
+  const redirectTo = await provider.interactionResult(req, res, results);
+  
+  res.send({ redirectTo });
+});
+
+// with koa
+router.post('/interaction/:grant', async (ctx, next) => {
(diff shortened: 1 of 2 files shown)


panva:
Hi @JSantha and thank you for this PR.

Is returning the interaction object the right call here? Shouldn't just the target url be returned instead?

Can you inline doc mark `setInteractionResult`, `interactionFinished`, `interactionDetails` and `setProviderSession` as public like so?

```
  /**
   * @name setProviderSession
   * @api public
   */
```

lib/provider.js
```diff
@@ -197,27 +197,43 @@ class Provider extends events.EventEmitter {
...
+  /**
+   * @name interactionDetails
+   * @api public
+   */
   async interactionDetails(req) {
     return getInteraction.call(this, req);
   }
 
+  /**
+   * @name interactionDetails
```
panva: should be `setProviderSession`

docs/configuration.md
```diff
@@ -320,6 +320,27 @@ router.post('/interaction/:grant', async (ctx, next) => {
 }
 ```
 
+**`#provider.interactionResult`**
+Unlike `#provider.interactionFinished` redirect uri is returned instead of immediate http redirect.
```
panva: replace `redirect uri` with `authorization request resume uri`

JSantha:
You are right with returning whole interaction, I've changed that to return `redirectTo` only. Renamed `setInteractionResult` to `interactionResult` as well to make naming more convenient.

panva:
Thank you @JSantha

Issue #352, opened 2018-08-13
Add metadata to setProviderSession


skeltonm:
**Is your feature request related to a problem? Please describe.**
I am implementing a password complexity check system during the login process, and I would like to add some metadata to `setProviderSession` that I can retrieve later when updating the user's password.

**Describe the solution you'd like**
Would it be possible to add a `meta` key to the object passed to `setProviderSession` that can store metadata?  Or would there be a better approach to doing what I am attempting to do?

 - [x] i have searched the configuration section for this feature and couldn't find it
 - [x] i have searched the issues tracker on github for similar requests and couldn't find anything related.

panva:
Sure, so long as the meta is scoped under a client. So that the result is inline with what `interactionResult` meta does.

Can you work out the PR for adding this and tests for it?

skeltonm:
Sure, I'll work on that.

panva:
i'm thinking something like this should be the signature

```js
async setProviderSession(req, res, {
  account, // account id string
  ts = epochTime(), // [optional] login timestamp, defaults to current timestamp
  remember = true, // [optional] set the session as persistent, defaults to true
  clients = [], // [optional] array of client id strings to pre-authorize in the updated session
  meta: { // [optional] object with keys being client_ids present in clients with their respective meta
    [client_id]: {},
  }
} = {})
```

PR #353, opened 2018-08-13
Add metadata to setProviderSession

skeltonm:
@panva This is more of a starting point so I can get your feedback on what needs to be updated or changed with the implementation of the meta argument.

Ref #352

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -363,13 +363,13 @@ async setProviderSession(req, res, {
 ```js
 // with express
 expressApp.post('/interaction/:grant/login', async (req, res) => {
-  await provider.setProviderSession(req, res, { account: 'accountId' }, { [client_id]: { key: 'value' } });
+  await provider.setProviderSession(req, res, { account: 'accountId' });
   // ...
 });
 
 // with koa
 router.post('/interaction/:grant/login', async (ctx, next) => {
-  await provider.setProviderSession(req, res, { account: 'accountId' }, { [client_id]: { key: 'value' } });
+  await provider.setProviderSession(req, res, { account: 'accountId' });
   // ...
 });
 ```
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -369,7 +369,7 @@ expressApp.post('/interaction/:grant/login', async (req, res) => {
 
 // with koa
 router.post('/interaction/:grant/login', async (ctx, next) => {
-  await provider.setProviderSession(req, res, { account: 'accountId' });
+  await provider.setProviderSession(ctx.req, ctx.res, { account: 'accountId' });
   // ...
 });
 ```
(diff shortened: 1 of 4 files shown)


docs/configuration.md
```diff
@@ -354,19 +354,22 @@ async setProviderSession(req, res, {
...
+    [client_id]: {},
+  }
 } = {})
 ```
 
 ```js
 // with express
 expressApp.post('/interaction/:grant/login', async (req, res) => {
-  await provider.setProviderSession(req, res, { account: 'accountId' });
+  await provider.setProviderSession(req, res, { account: 'accountId' }, { [client_id]: { key: 'value' } });
```
panva: add the clients array to this example then, or omit the meta from it.

docs/configuration.md
```diff
@@ -354,19 +354,22 @@ async setProviderSession(req, res, {
...
   // ...
 });
 
 // with koa
 router.post('/interaction/:grant/login', async (ctx, next) => {
-  await provider.setProviderSession(ctx.req, ctx.res, { account: 'accountId' });
+  await provider.setProviderSession(req, res, { account: 'accountId' }, { [client_id]: { key: 'value' } });
```
panva: dtto.

lib/provider.js
```diff
@@ -248,6 +249,7 @@ class Provider extends events.EventEmitter {
     Object.assign(session, {
       account,
       loginTs: ts,
+      meta,
```
panva: Nope, similar to `sidFor` it should call `metaFor` with the provided client_id and meta object. It should also assert that the client_id was provided in the clients array.

lib/models/session.js
```diff
@@ -40,6 +40,10 @@ module.exports = function getSession(provider) {
       return this.loginTs;
     }
 
+    meta() {
```
panva: this is not necessary, there's `metaFor`

panva:
@skeltonm thank you for proposing this, i left my comments in the review.

docs/configuration.md
skeltonm: I removed the metadata since I put it in the wrong spot anyway.

lib/models/session.js
skeltonm: Removed

docs/configuration.md
```diff
@@ -366,7 +369,7 @@ expressApp.post('/interaction/:grant/login', async (req, res) => {
 
 // with koa
 router.post('/interaction/:grant/login', async (ctx, next) => {
-  await provider.setProviderSession(ctx.req, ctx.res, { account: 'accountId' });
+  await provider.setProviderSession(req, res, { account: 'accountId' });
```
panva: ctx.req, ctx.res got changed to req, res

skeltonm: Ah, good catch.  I did a copy and paste from the Express example.

skeltonm: Updated

lib/provider.js
```diff
@@ -249,13 +249,17 @@ class Provider extends events.EventEmitter {
...
       account,
       loginTs: ts,
-      meta,
     });
     if (!remember) session.transient = true;
     clients.forEach((clientId) => {
       assert(typeof clientId === 'string', 'clients must contain an array of client_id strings');
       session.sidFor(clientId, uuid());
     });
+    Object.keys(meta).forEach((clientId) => {
```
panva: please use `Object.entries(meta).forEach(([clientId, meta]) => { ... });` instead

PR #384, opened 2018-10-19
fix: allow http2 in assertReqRes (fixes #383)

### LINKED ISSUE #383: HTTP/2 Compatibility
**Describe the bug**
The provider expects an http or https request and response object in [assertReqRes(req, res)](https://github.com/panva/node-oidc-provider/blob/531dfbf64a4d9a3dd7e5403f211fab4d4a590941/lib/provider.js#L37-L48). This will throw an AssertionError if an [http2](https://nodejs.org/api/http2.html#http2_compatibility_api) request and response object is provided. But Koa supports the Compatibility API of http2 (see [koajs/koa#1193](https://github.com/koajs/koa/issues/1193#issuecomment-389713217)). And I did not notice any problems with the provider either.

**To Reproduce**
Replace the startup of the server:
```js
const https = require('https');
// ...
const server = https.createServer(tlsOptions, provider.app.callback());
// ...
```
... with the following HTTP/2 startup:
```js
const https2 = require('https2');
// ...
const http2Wrap = callback => (req, res) => {
    // fix to koa/request.host
    req.headers.host = req.headers.host || req.headers[':authority'];
    callback(req, res);
};

const server = http2.createSecureServer({
    allowHTTP1: true, // backward compatibility
    ...tlsOptions,
}, http2Wrap(provider.app.callback()));
// ...
```

Steps to reproduce the behaviour:
1.  Start the provider.
2. Start an interaction.
3. `getInteraction` will call `assertReqRes` and an AssertionError is thrown

**Expected behaviour**
No AssertionError 😃

**Environment:**
 - oidc-provider version: v5.1.0
 - node version: v10.11.0

**Possible solution:**
Change `assertReqRes`:
```js
function assertReqRes(req, res) {
  assert(
    req instanceof http.IncomingMessage || req instanceof http2.Http2ServerRequest,
    'first argument must be the request (http.IncomingMessage), for express req, for koa ctx.req',
  );
  if (arguments.length === 2) {
    assert(
      res instanceof http.ServerResponse || res instanceof http2.Http2ServerResponse,
      'second argument must be the response (http.ServerResponse), for express res, for koa ctx.res',
    );
  }
}
```

**Additional context**

 - [x] the bug is happening on latest oidc-provider too.
 - [x] i have tried DEBUG (see readme.md) and can see the issue is with the provider and not my code.
 - [x] i have searched the issues tracker on github for similar issues and couldn't find anything related.
 - [x] i have searched the [FAQ](https://github.com/panva/node-oidc-provider/wiki/FAQ) and couldn't find anything related.

mgjm:
This is the fix for #383 and adds support for http2 in `assertReqRes`.

The http2 module is only stable in node >= v10.0.0. Therefore http2 is only supported if node >= v10.0.0 is used.

### DIFF
--- a/lib/provider.js
+++ b/lib/provider.js
@@ -1,7 +1,10 @@
+const HTTP2_SUPPORTED = Number(process.version.match(/^v(\d+)/)[1]) >= 10;
+
 const url = require('url');
 const assert = require('assert');
 const http = require('http');
 const events = require('events');
+const http2 = HTTP2_SUPPORTED ? require('http2') : null;
 
 const { get, merge } = require('lodash');
 const Koa = require('koa');
@@ -36,12 +39,14 @@ function getCookie(name, opts, cookies) {
 
 function assertReqRes(req, res) {
   assert(
-    req instanceof http.IncomingMessage,
+    req instanceof http.IncomingMessage
+    || (HTTP2_SUPPORTED && req instanceof http2.Http2ServerRequest),
     'first argument must be the request (http.IncomingMessage), for express req, for koa ctx.req',
   );
   if (arguments.length === 2) {
     assert(
-      res instanceof http.ServerResponse,
+      res instanceof http.ServerResponse
+      || (HTTP2_SUPPORTED && res instanceof http2.Http2ServerResponse),
       'second argument must be the response (http.ServerResponse), for express res, for koa ctx.res',
     );
   }
--- a/lib/provider.js
+++ b/lib/provider.js
@@ -1,17 +1,23 @@
-const HTTP2_SUPPORTED = Number(process.version.match(/^v(\d+)/)[1]) >= 10;
-
 const url = require('url');
 const assert = require('assert');
-const http = require('http');
+const { IncomingMessage, ServerResponse } = require('http');
 const events = require('events');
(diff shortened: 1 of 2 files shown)


mgjm:
I just noticed that http2 did not reach stable in node `v10.0.0` but in `v10.10.0`. II will update the PR later.

panva:
I actually wouldn't check for it being stable, just try / catch the `require('http2')`, that way we cover lts/8 experimental as well as those relying on deprecated `http2` package.

mgjm:
The problem with this approach is that it will always trigger an `ExperimentalWarning` in node < `v10.10.0`.

panva:
good point 🤔

panva:
> Edit: Btw. http2 is available (as experimental) since v8.4.0 so no try / catch would be required because node-oidc-provider requires node >= v8.9.0.

figured that too, but i get the point about experimental warning. added a semver check in 841ddb2

mgjm:
Nice, this seems like a good way.

I also just submitted a PR to koa to fix the host property for http2. So that the following hack will not be needed anymore:
```js
const http2Wrap = callback => (req, res) => {
    // fix to koa/request.host
    req.headers.host = req.headers.host || req.headers[':authority'];
    callback(req, res);
};
```

PR #389, opened 2018-10-29
feat: add Typescript declarations

javlund:
First version of the declarations file. Please comment or, even better, add commits!

Note that I'm honestly somewhat a rookie in Typescript, especially when it comes to declaration files and the whole build aspect of it, so I'd really like more eyes on this.

What I've done so far:
* added the declarations file
* added a `tsconfig.json` file (not sure if needed?)
* as a test, run `tsc -p .` in the root dir and then tried starting up a new file importing the provider. Noted that types were available.


panva:
not a TS guy at all, but i need to get into it to at least maintain the bundled types after this.

So take my feedback with some reserves

- `.gitignore` change is not needed, we won't be generating any new files, this will remain a js source library, just with bundled types
- same for `tsconfig.json` - there's no compilation, is there?
- we need a "types" or "typings" entry in `package.json` tho

question: `index.d.ts` can this be an actual index file that imports types from other `d.ts` files throughout the codebase? In the end i'd split this up by components or interfaces.
question2: other than types, could the file also contain descriptions for functions, arguments, etc. that they show up in people's IDEs?

I will go through the proposed types and comment inline on what's wrong / incomplete and also what's missing from the public API or what's private and shouldn't be typed.

index.d.ts
```diff
@@ -0,0 +1,456 @@
+declare module 'oidc-provider' {
```
panva: 2 spaces, not 4 for the whole file please

index.d.ts
```diff
@@ -0,0 +1,456 @@
...
+    import { Request, Response, Handler } from 'express';
+    import Koa = require('koa');
+    import Router = require('koa-router');
+    import { IncomingMessage, ServerResponse } from 'http';
+    import EventEmitter = require('events');
+    import { JWK } from 'node-jose';
```
panva: iirc the types for node-jose are in an awful state, what would it mean if we were not to import them?

index.d.ts
```diff
@@ -0,0 +1,456 @@
...
+
+    type SubjectType = 'public' | 'parwise';
+
+    interface IClient {
```
panva: awful lot of metadata properties missing, please combine the following two sources

1) generic metadata https://github.com/panva/node-oidc-provider/blob/v5.1.2/lib/consts/client_attributes.js#L1-L28
2) feature specific metadata (`RECOGNIZED_METADATA.push`) from https://github.com/panva/node-oidc-provider/blob/v5.1.2/lib/helpers/client_schema.js#L53-L126

index.d.ts
```diff
@@ -0,0 +1,456 @@
...
+        applicationType?: 'web' | 'native';
```
panva: is there syntax to specify the default value? if so `web` is default.

index.d.ts
```diff
@@ -0,0 +1,456 @@
...
+        grant_types?: string[];
```
panva: grant_types is not optional, an empty array is the minimum accepted value.

index.d.ts
```diff
@@ -0,0 +1,456 @@
...
+        id_token_signed_response_alg?: string;
```
panva: defaults to RS256

index.d.ts
```diff
@@ -0,0 +1,456 @@
...
+        require_auth_time?: boolean;
```
panva: defaults to false

index.d.ts
```diff
@@ -0,0 +1,456 @@
...
+        response_types?: string[];
```
panva: response_types is not optional, an empty array is the minimum accepted value

index.d.ts
```diff
@@ -0,0 +1,456 @@
...
+        subject_type?: SubjectType;
+        token_endpoint_auth_method?: string;
```
panva: defaults to `client_secret_basic` and there's an enum list for the supported values. dtto for introspection and revocation counterparts

index.d.ts
```diff
@@ -0,0 +1,456 @@
...
+        request_uris?: string[];
+        client_id?: string;
```
panva: client_id is not optional

index.d.ts
```diff
@@ -0,0 +1,456 @@
...
+        client_secret?: string;
+        redirect_uris: string[];
+        introspection_endpoint_auth_method?: string;
+    }
+
+    interface IAdapter {
+        upsert: (id: string, payload: string, expiresIn: number) => Promise<void>;
```
panva: payload is an object

index.d.ts
```diff
@@ -0,0 +1,456 @@
...
+        find: (id: string) => Promise<void>;
```
panva: promise should resolve an object previously stored in upsert (payload argument)

index.d.ts
```diff
@@ -0,0 +1,456 @@
...
+        findByUserCode: (userCode: string) => Promise<void>;
```
panva: promise should resolve an object

index.d.ts
```diff
@@ -0,0 +1,456 @@
...
+        consume: (id: string) => Promise<void>;
+        destroy: (id: string) => Promise<void>;
+        connect?: (provider: Provider) => Promise<void>;
+        key: (id: string) => string;
```
panva: remove this, that's not part of the expected API, some example adapters just use this internally

index.d.ts
```diff
@@ -0,0 +1,456 @@
...
+    }
+
+    interface IKeystore {
```
panva: IKeystore is unused, but i'm wondering if it's gonna be better to maintain own JWK.KeyStore types here.

ulrikstrid:
Great that this is worked on! As this is based on the typings I created I will say that they were created as the minimal working thing that gave me some help.

Another approach for incrementally adding TypeScript in to the codebase is to add jsdoc comments and generate `.d.ts` from that, is that something you would be open to @panva? I think that would have a better chance of staying up-to-date.

panva:
> Another approach for incrementally adding TypeScript in to the codebase is to add jsdoc comments and generate .d.ts from that, is that something you would be open to @panva? I think that would have a better chance of staying up-to-date.

I'm very interested in this, would you be willing to work out the necessary tooling for this? And a small example of what's possible with this?

ulrikstrid:
I won't be able to do it now but might have some time in a couple of weeks as I'm in the process of changing employer.

But if anyone else wants to pick it up in the meantime here are some resources that come to mind:
An example of a library that used jsdoc is this https://github.com/Azure/azure-storage-node.
And according to the docs it seems like it should "just work" https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html

index.d.ts
javlund: No, interfaces cannot have default values as I understand it. We could perhaps convert it to a class instead?

index.d.ts
javlund: Same issue with interfaces not being able to have default values - convert to class?

javlund:
I have pushed some commits addressing a number of your comments @panva (but not all), feel free to revisit.

Regarding your questions:
1) I believe so, according to https://www.typescriptlang.org/docs/handbook/declaration-merging.html
2) As I understand it that is done through JSDoc, which seems to already be on the table now.

index.d.ts
javlund: In my opinion, the less we import the better as this might require us to state the additional types required in our `package.json` `devDependencies`.

index.d.ts
javlund: Can you provide suggested types for them? Then I'd love to comb through the files and include the ones that aren't already there.

javlund: I ended up trying to put all of them in in an additional commit. Take a look and let me know if it seems sensible.

index.d.ts
javlund: I'm thinking that would make sense. Could you give me an example structure?

index.d.ts
panva: what would be the implication of converting to class?

index.d.ts
panva: I'm thinking the types can just specify a jwks object as an interface, e.g. 

```js
{
  keys: [
    rsaKey | ecKey
    , ...
  ]
};

rsaKey = {
  kty: 'RSA',
  d: required string
  dp: required string
  dq: required string
  e: required string
  n: required string
  p: required string
  q: required string
  qi: required string
  use: optional string 'sig' | 'enc'
  kid: optional string,
  alg: optional string,
}

ecKey = {
  kty: 'EC',
  crv: required string
  x: required string
  y: required string
  d: required string
  use: optional string 'sig' | 'enc'
  kid: optional string,
  alg: optional string,
}
```

i believe most people use it this way anyway, passing the jose jwk keystore is just a QoL change.

index.d.ts
```diff
@@ -0,0 +1,355 @@
...
+    interactionDetails(ctx: IncomingMessage): Promise<ISession>;
+
+    setProviderSession(req: IncomingMessage, res: ServerResponse, {}): Promise<ISession>;
+
+    interactionFinished(req: IncomingMessage, res: ServerResponse, {}): Promise<void>;
+
+    callback: Handler;
+    listen(port: string | number): void;
+
+    Client: IClient;
```
panva: i'd say minimal types don't need to type classes that are used for advanced customization, so please remove Client from here, same for `session`

index.d.ts
```diff
@@ -0,0 +1,355 @@
...
+
+    session: ISession;
+
+    params: {
```
panva: both params and result don't belong here, i guess a mistake, please remove

index.d.ts
```diff
@@ -0,0 +1,355 @@
...
+      response_type: "none";
+    };
+    result: boolean;
+  }
+
+  export const createKeyStore: JWK.createKeyStore;
```
panva: let's remove these JWK exports and remove the JWK import

index.d.ts
```diff
@@ -0,0 +1,355 @@
...
+  export const asKeyStore: JWK.asKeyStore;
+  export const asKey: JWK.asKey;
+
+  export default Provider;
```
panva: as a minimal type implementation let's type the exported errors property

see:
- https://github.com/panva/node-oidc-provider/blob/v5.1.2/test/provider/provider_class.test.js#L12-L41
- https://github.com/panva/node-oidc-provider/blob/v5.1.2/lib/helpers/errors.js

panva:
@javlund many thanks for your efforts!

javlund:
Hi @panva , sorry for my absence, a conference and family life required my attention for a few days. I've dealt with most of the things you've brought up, however haven't looked into the errors part yet. Not 100% on to the best way to define those as they seem to dictate restraints on the object key names. Maybe I'm just a bit too tired right now.

panva:
Sorry for the lack of feedback, i'm really short on time lately.

javlund:
I'm thinking we might be able to merge this as it is and then let the errors come in in a future commit?

panva:
> I'm thinking we might be able to merge this as it is and then let the errors come in in a future commit?

I'm not comfortable doing that. Is anyone able to do a spike for generating the typings from inline docs?

ulrikstrid:
After doing some more research on this this seems to be the best bet right now: https://github.com/englercj/tsd-jsdoc#readme
Another solution *might* be to just use `alloJS` flag in consumer land and that should parse the types but this doesn't seem like a longterm solution.

panva:
I don’t think as-is the types are complete and i don’t have the capacity to align them. Furthermore they’re not future proof with the next major WIP.

Of course there’s nothing stopping you from publishing to definitely-typed but i feel you’d be making the rest of the ts developers a disservice with types that don’t cover the full feature set and/or are outright incorrect.

I’ma close this and leave the discussion open but unless I fall for TS myself I won’t be bundling it with the lib afterall.

javlund:
I wouldn't dream of releasing them without your consent.

Would it make sense to reopen the discussion when next major version is released?

panva:
> I wouldn't dream of releasing them without your consent.

I can't control what the TS community maintains for itself. Just wondering, if someone releases crap types, what are the mechanisms to curate and fix those?

theaccordance:
@javlund did you end up publishing the definitions for this repo to DefinitelyTyped?  I'm having difficulty sourcing the typed file.  

If you didn't, how much additional work is required to continue with this definition file?

panva:
Releasing any incomplete types for a project this size and flexibility would be a disservice to anyone unfortunate enough to use them.

theaccordance:
Not having a Definition file is doing a disservice to my efforts to incorporate this dependency into my application.  

I know you're busy with 6.x which is why I didn't include you in my mention; I'm just looking for the latest version of the type definitions, and some understanding of its state/progress, that way I can contribute if I encounter gaps

javlund:
My latest effort is available for anyone to see in this PR. As @panva wasn't satisfied I saw no point in publishing them.

Personally, I strongly believe this project should contain TS definitions, but they should be sanctioned by the creator and focus on versions 6+ instead of a version that is approaching EOL.

panva:
@javlund thanks, i'm fine revisiting this when v6.0.0 goes out, i'm counting on your support ;)

javlund:
@panva Certainly!

panva:
After writing `.d.ts` typescript definitions for openid-client, @panva/jose, and paseto I took a stab at doing so for `oidc-provider`.

I kindly ask everyone to try them and provide feedback, 👍 or 👎 in #530

panva:
Typings released in v6.9.0

PR #393, opened 2018-11-02
refactor: enhance interactionResult function

supermanHub:
In order to reduce additional query effort for interaction details, I would like to expose much more information in `interactionResult` function, typically, the entire interaction result!

On the other hand, returning the entire interaction result is more consistent with the name of the method. after all, `interaction result` not only contains `returnTo`.

BTW: the `interactionResult` will do the `session` check as well.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -377,23 +377,24 @@ router.post('/interaction/:grant', async (ctx, next) => {
 ```
 
 **`#provider.interactionResult`**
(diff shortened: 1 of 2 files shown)


panva:
Please see the contribution guideline. Unless previously agreed upon (i.e. fixing a bug) all contributions must be backwards compatible. 

This isnt. I can stash it and release once there’s a need for v6 or you refactor in a backwards compatible way.

Issue #415, opened 2018-12-18
Make keystore.refresh() safer


davidgtonge:
**Is your feature request related to a problem? Please describe.**
Refreshing a keystore for a client that has registered with a set of keys rather than a `jwks_uri` throws an unrelated error.

**Describe the solution you'd like**
While debugging this problem I noticed that all instances of `keystore.refresh()` first checked if the keystore was stale. I will submit a PR that pulls that check into the refresh method itself.

**Describe alternatives you've considered**
Documentation? Or better error messages? Or an explicit jwks_uri check in the refresh method?

**Additional context**

 - [*] i have searched the configuration section for this feature and couldn't find it
 - [*] i have searched the issues tracker on github for similar requests and couldn't find anything related.

davidgtonge:
Thanks @panva

PR #416, opened 2018-12-18
feat: make refreshing a keystore safer

davidgtonge:
This move the staleness check (which contains a
check if a jwks_uri exists) into the refresh method.
This reduces some duplication and makes it easier
to use for people who interact with `client.keystore`.

I added an option to force a refresh, as this was
needed to enable the tests to pass.

#415

### DIFF
--- a/lib/helpers/jwt.js
+++ b/lib/helpers/jwt.js
@@ -2,10 +2,16 @@ const assert = require('assert');
 
 const base64url = require('base64url');
 const {
-  JWK: { isKeyStore },
+  JWK: { createKeyStore },
   JWS: { createSign, createVerify },
   JWE: { createEncrypt, createDecrypt },
 } = require('node-jose');
+
+const JWKStore = createKeyStore().constructor;
+
+function isKeyStore(obj) {
+  return obj instanceof JWKStore;
+}
 
 const epochTime = require('./epoch_time');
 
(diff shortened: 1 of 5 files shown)


panva:
Isn't this opening up an uncaught if you use `refresh: true` on a keystore that does not have a `jwksUri`?

lib/helpers/jwt.js
```diff
@@ -108,7 +108,7 @@ class JWT {
     try {
       verified = await createVerify(keyOrStore).verify(jwt);
     } catch (err) {
-      if (isKeyStore(keyOrStore) && keyOrStore.stale && keyOrStore.stale()) {
+      if (isKeyStore(keyOrStore)) {
```
panva: I think we at least need a check for the `refresh` function being present on the keyOrStore instance. While i dislike overloading with the node-jose prototype i don't have the time to refactor this to use an internal representation of a keystore :(

davidgtonge:
> Isn't this opening up an uncaught if you use `refresh: true` on a keystore that does not have a `jwksUri`?
> 
> If you don't need the `refresh` argument then i suggest to remove it and simply mock response of keystore's `fresh()` for those tests that are testing the refresh functionality.

I agree that mocking the `fresh()` test is better. I had a go using Sinon but was struggling with stubbing out the fresh function, e.g.:

`sinon.stub(this.client.keystore, 'fresh').returns(false);` works fine

but I get errors when calling:

`this.client.keystore.fresh.restore();`
`Cannot assign to read only property 'fresh' of object '#<JWKStore>'`

I don't have much experience mocking out functions that are defined using `Object.defineProperties`, do you know a way around this?

panva:
I'll have a go at this then.

panva:
wdyt?

Issue #417, opened 2018-12-19
adapters mongodb deleteMany is not a function


huangyingwen:
**Describe the bug**

![image](https://user-images.githubusercontent.com/1214875/50196544-78dca700-037e-11e9-8b0e-f20b24b059b8.png)

```
  async destroy(_id) {
    const coll = await this.coll();
    const found = await coll.findOneAndDelete({ _id });
    if (found.value && found.value.grantId) {
      const promises = [];

      collections.forEach(name => {
        if (grantable.has(name)) {
          promises.push(this.coll(name).deleteMany({ grantId: found.value.grantId }));
        }
      });

      await Promise.all(promises);
    }
  }
```

this.coll(name) are asynchronous, In the code, the call is synchronous

**To Reproduce**
Provider and Client configuration: (inline or gist)
```js
// provider configuration

// keystore(s) - if they affect the behaviour

// client(s) configuration

```

Steps to reproduce the behaviour:

1.  
2.  
3.  

**Expected behaviour**

**Environment:**
 - oidc-provider version: [e.g. v4.3.0]
 - node version: [e.g. v8.9.0]

**Additional context**

 - [x] the bug is happening on latest oidc-provider too.
 - [x] i have tried DEBUG (see readme.md) and can see the issue is with the provider and not my code.
 - [x] i have searched the issues tracker on github for similar issues and couldn't find anything related.
 - [x] i have searched the [FAQ](https://github.com/panva/node-oidc-provider/wiki/FAQ) and couldn't find anything related.

**Logs**

panva:
Hi @huangyingwen,

I am not able to reproduce the behaviour, `DB.collection` without a callback and flag for strict mode is synchronous.

Are you sure you're using the latest driver?

Please provide steps to reproduce using the example app or bundle up one of your own that i can replicate the issue locally.

I will close this and the PR for now but feel free to post the details and we'll reopen with a proper fix should there really be an issue.

PR #418, opened 2018-12-19
fix: adapters mongodb deleteMany is not a function

huangyingwen:
adapters mongodb deleteMany is not a function
https://github.com/panva/node-oidc-provider/issues/417


panva:
https://github.com/panva/node-oidc-provider/issues/417#issuecomment-448517695

PR #445, opened 2019-03-31
Inform developer that adapter property should be a class constructor

big-kahuna-burger:


lib/helpers/initialize_adapter.js
```diff
@@ -7,6 +7,15 @@ module.exports = async function initializeAdapter(adapter = MemoryAdapter) {
   if (adapter === MemoryAdapter) {
     attention.warn('a quick start development-only MemoryAdapter is used');
   }
+  function isCtor() {
+    return !!adapter.prototype && !!adapter.prototype.constructor.name;
+  }
+
+  if (!isCtor()) {
+    throw new Error(
+      "Expected 'adapter' to be a constructor, provide a valid adapter in Provider config.",
```
panva: Please swap the " and ', `'` as outer, `"` as inner around the word adapter.

lib/helpers/initialize_adapter.js
```diff
@@ -7,6 +7,15 @@ module.exports = async function initializeAdapter(adapter = MemoryAdapter) {
   if (adapter === MemoryAdapter) {
     attention.warn('a quick start development-only MemoryAdapter is used');
   }
+  function isCtor() {
```
panva: creating a function like this is not necessary, just use the expressions in the if.

lib/helpers/initialize_adapter.js
```diff
@@ -7,6 +7,15 @@ module.exports = async function initializeAdapter(adapter = MemoryAdapter) {
...
+    return !!adapter.prototype && !!adapter.prototype.constructor.name;
```
panva: no need to double `!` either of these wdyt?

big-kahuna-burger:
see latest commit

panva:
@big-kahuna-burger thank you for the contribution

PR #455, opened 2019-04-17
Passed ctx into id factory

jaxoncreed:
@panva Do you think that this change would be in the spirit of the openid standards?

Here's the use case: I would like to generate client_ids based on the Origin header during dynamic registration. I'll then use the client_id to validate if provided redirect URLs are valid (only redirect urls with the same hostname as the client_id will be allowed). I know that "it’s best that it isn’t guessable by third parties" (https://www.oauth.com/oauth2-servers/client-registration/client-id-secret/), but putting the extra restriction on the redirect should negate that weakness.

For more information on the reasoning behind this, see this thread (https://github.com/solid/webid-oidc-spec/issues/12#issuecomment-481835310)

### DIFF
--- a/lib/actions/registration.js
+++ b/lib/actions/registration.js
@@ -91,7 +91,7 @@ module.exports = {
       const { oidc: { provider } } = ctx;
       const { idFactory, secretFactory } = instance(provider).configuration('features.registration');
       const properties = {};
-      const clientId = String(idFactory());
+      const clientId = String(idFactory(ctx));
 
       const rat = new provider.RegistrationAccessToken({ clientId });
       ctx.oidc.entity('RegistrationAccessToken', rat);


panva:
Not being familiar with WebID i think it shouldn't go the route of dynamic registrations for what are essentially throw-away clients, aren't they? Is there a point in keeping the client around when the end-user authenticates? It could take inspiration from [Self-Issued OpenID Provider](https://openid.net/specs/openid-connect-core-1_0.html#SelfIssued). A client registration is done on the fly as part of the authorization request with the `registration` parameter.

I could be wrong tho.

The change itself I don't have a problem with but it looks smelly, like a one-off serving that exact use case and none other so i'll go ahead and close while trying to think of something general, e.g. not having the factories called from the registration route but rather the Client model itself when it already has all the other metadata assigned.

jaxoncreed:
@panva Thanks for pointing out the self-issued OpenID provider specs. I think that's probably what we want to follow. Do you think my update makes sense for this use case?

PR #469, opened 2019-05-24
fix: add scope to implicit responses when different from request

ayZagen:
`scope` in authorization response must exist if it is not equal to the one requested by client. 

>   The authorization server MAY fully or partially ignore the scope
>    requested by the client, based on the authorization server policy or
>    the resource owner's instructions.  If the issued access token scope
>    is different from the one requested by the client, the authorization
>    server MUST include the "scope" response parameter to inform the
>    client of the actual scope granted.
See [RFC6749](https://tools.ietf.org/html/rfc6749#section-3.3)

It is stated as optional in [Section 5.1 - Issuing an Access Token - Successful Response](https://tools.ietf.org/html/rfc6749#section-5.1)

>    scope
>    &nbsp;&nbsp;&nbsp;&nbsp;OPTIONAL, if identical to the scope requested by the client;
>    &nbsp;&nbsp;&nbsp;&nbsp;otherwise, REQUIRED.  The scope of the access token as
>    &nbsp;&nbsp;&nbsp;&nbsp;described by Section 3.3.

Also you could see `scope` is optional in token responses for all OAuth2 grant types.
See followings:
[Authorization Code Grant](https://tools.ietf.org/html/rfc6749#section-4.1.4)
[Implicit Grant](https://tools.ietf.org/html/rfc6749#section-4.2.2)
[Resource Owner Password Credentials Grant](https://tools.ietf.org/html/rfc6749#section-4.3.3)
[Client Credentials Grant](https://tools.ietf.org/html/rfc6749#section-4.4.3)

### DIFF
--- a/lib/actions/authorization/process_response_types.js
+++ b/lib/actions/authorization/process_response_types.js
@@ -28,11 +28,17 @@ async function tokenHandler(ctx) {
 
   ctx.oidc.entity('AccessToken', token);
 
-  return {
+  const result = {
     access_token: await token.save(),
     expires_in: token.expiration,
     token_type: 'Bearer',
   };
+
+  if (token.scope !== ctx.oidc.params.scope) {
+    result.scope = token.scope;
+  }
+
+  return result;
 }
 
 async function codeHandler(ctx) {
--- a/test/core/hybrid/code+id_token+token.authorization.test.js
+++ b/test/core/hybrid/code+id_token+token.authorization.test.js
@@ -105,6 +105,30 @@ describe('HYBRID code+id_token+token', () => {
       });
     });
 
+    describe(`${verb} ${route} with the scope not being fulfilled`, () => {
+      before(function () {
+        return this.login({
+          scope: 'openid profile email',
+          rejectedScopes: ['email'],
+        });
+      });
+
+      it('responds with an extra parameter scope', async function () {
+        const auth = new this.AuthorizationRequest({
+          response_type,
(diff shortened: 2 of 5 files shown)


ayZagen:
@panva But I am not sure that we should compare them order regardless

ayZagen:
> i believe if you're implementing a new system you should not use response types issuing access tokens via the front channel

The response json is generated there so I thought that is the perfect place to do such check and implementation. 

> I'm adding the scope to the response regardless of the original one for token response. 

Adding it regardless of equality would create a high payload because scopes could be really large. I believe if granted and requested ones are equal, no need to increase the size of response.

> That's simply because I have no way of knowing whether they've been crossed off before issuing the authorization code

`ctx.oidc.params` contains the raw parameters of authorization request and `process_response_type` is could be counted as the final step of token generating -_maybe response_mode handling could be counted but this place is more relevant in my opinion_-. So, I believe checking and mutating the result there would be appropriate for your concern.

> Changes requested: Please add tests
I was going to implement test however I need a kind of hook for modifying the scopes and didn't want to pollute the codebase without your confirmation. I will submit my code as soon as I have spare time for it.

panva:
> I was going to implement test however I need a kind of hook for modifying the scopes and didn't want to pollute the codebase without your confirmation. I will submit my code as soon as I have spare time for it.

No hooks are necessary, you can simply simulate interactions that has a scope already rejected.

Don't worry about the tests i'll take this over now.

PR #474, opened 2019-05-28
Fixed enabled check conditions

thomasvargiu:
Fixed enabled check conditions in `discovery.js`


panva:
@thomasvargiu thanks, i've missed this.

PR #480, opened 2019-06-05
Fix jwtResponseModes draft version

DarkSorrow:
The version of this draft is 02 but the version being acknowledged was the 01


panva:
Duplicate of #477 where i've explained what needs to happen, the draft update was purely fixing typos, not changing normative requirements. Acked 1 and 2 should both work and that's what i've requested be changed in #477

Issue #466, opened 2019-05-21
sub missing in id_token when dynamicScopes are used


ayZagen:
**Describe the bug**
Setting dynamicScopes to `/.*/` -which means accept all scopes- makes sub excluded from id_token. However sub in id_token is a must according to [OpenId Specification ](https://openid.net/specs/openid-connect-core-1_0.html#IDToken)

**To Reproduce**
set dynamicScopes to accept all as following:

```js
dynamicScopes: [ /.*/ ]

```

version: 6.0.0-alpha.6 (but i believe it is the same on v5)

panva:
Well, are you requesting the openid scope in the request?

ayZagen:
@panva Of course, this is the request:

URL:
http://localhost3000/oauth/auth

Query String Parameters:
client_id: test
redirect_uri: http://localhost:8080/callback
response_type: id_token token
scope: openid email
state: b0496cfedbbb49699ff15f9ee4d5092f
nonce: b1b53884cb37474aa4fd987d69bef2d1
audience: test-audience

panva:
yeah, i've confirmed this.

panva:
Thank you for the report.

ayZagen:
Thanks for the great work you have done!

PR #483, opened 2019-06-21
FIX: sub not present in id_token.

codestorycooked:
When using dynimic scopes as outlined here https://github.com/panva/node-oidc-provider/issues/466 id_token does not contains `sub` field. This is already fixed in 6.X version and this is for 5.X version


codestorycooked:
Hi @panva  how can I provide this as 5.3.1? I am not sure how to resolve this conflict

PR #486, opened 2019-06-24
fix: use fixed scope to claim mapping over dynamic ones

codestorycooked:
When using dynimic scopes as outlined here #466 id_token does not contains sub field. This is already fixed in 6.X version and this is for 5.X version


codestorycooked:
> When using dynimic scopes as outlined here #466 id_token does not contains sub field. This is already fixed in 6.X version and this is for 5.X version

I as per your review, I created PR based from 5.X. Let me know if this is good.

PR #509, opened 2019-08-07
idTokenSigningAlgValues uses whitelistedJWA

jaxoncreed:
Before this used an `idTokenSigningAlgValues` from the root configuration. I'm not exactly sure what this was generated by, but it's value was only `['HS256']`. This caused problems for my RP which registered without an `id_token_signing_alg_values` and defaulted to `RS256`


panva:
> I'm not exactly sure what this was generated by

It is generated by algorithms available on the asymmetrical keys available in the `jwks` configuration property and it uses whitelistJWA, well, as a whitelist :)

This doesn't need fixing but i think your jwks configuration is not right and doesn't include a RS256 usable key.

Please open an issue instead, you should have anyway, this doesn't need fixing.

PR #511, opened 2019-08-12
Defer endSession form creation to logoutSource

fauxsoup:
I use react and do most of my rendering client side, so it's kind of a pain to have to render a page with an HTML string. I'd rather be given the components I need and create the form myself. I feel like this provides more flexibility.

This is just my feeling and I won't be offended if you prefer it the other way. This seemed sensible to me, though, so I thought I'd make a PR.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -2543,25 +2543,32 @@ HTML source rendered when session management feature renders a confirmation prom
 
 _**default value**_:
 ```js
-async logoutSource(ctx, form) {
+async logoutSource(ctx, action, secret) {
   // @param ctx - koa request context
-  // @param form - form source (id="op.logoutForm") to be embedded in the page and submitted by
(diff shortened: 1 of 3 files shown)


panva:
@fauxsoup the reason why this interface is as such kind of doesn't hold any more so i'm agreeing this change would be useful and more flexible.

That being said, its a breaking change and I cannot include it in the current track, i'll label the issue next-major and close it, the label will allow me to recall these kinds of proposals when we get to a next major.

Also, you can already render the whole html today given that you have access to the `ctx` and can just extract the action and secret from the passed in argument with a regexp.

Sounds good?

fauxsoup:
Yup! That was my plan if this wasn't accepted anyway, but I figured others could benefit from this too. 

Thank you for your continued work! 6.x is looking great, and I look forward to seeing the next major release.

PR #530, opened 2019-09-11
feat: added TypeScript definitions

panva:
After writing `.d.ts` typescript definitions for openid-client, @panva/jose, and paseto I took a stab at doing so for `oidc-provider`.

I kindly ask everyone to try them and provide feedback 👍 or 👎

Install with

```console
npm i panva/node-oidc-provider#ts
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ oidc-provider allows to be extended and configured in various ways to fit a vari
 the [available configuration](/docs/README.md).
 
 ```js
-const Provider = require('oidc-provider');
+const { Provider } = require('oidc-provider');
 const configuration = {
   // ... see available options /docs
   clients: [{
@@ -121,6 +121,33 @@ oidc.app
 
 // or just expose a server standalone, see /examples/standalone.js
 const server = oidc.listen(3000, () => {
+  console.log('oidc-provider listening on port 3000, check http://localhost:3000/.well-known/openid-configuration');
+});
+```
+
+```ts
+import * as oidc from 'oidc-provider';
+
+const configuration = {
+  // ... see available options /docs
+  clients: [{
+    client_id: 'foo',
+    client_secret: 'bar',
+    redirect_uris: ['http://lvh.me:8080/cb'],
+    // + other client properties
+  }],
+};
+
+const provider = new oidc.Provider('http://localhost:3000', configuration);
+
+// express/nodejs style application callback (req, res, next) for use with express apps, see /examples/express.js
+provider.callback
+
+// koa application for use with koa apps, see /examples/koa.js
+provider.app
+
+// or just expose a server standalone, see /examples/standalone.js
+const server = provider.listen(3000, () => {
   console.log('oidc-provider listening on port 3000, check http://localhost:3000/.well-known/openid-configuration');
 });
 ```
--- a/certification/docker.js
+++ b/certification/docker.js
@@ -6,7 +6,7 @@ const https = require('https');
 const pem = require('https-pem');
 const render = require('koa-ejs');
 
-const Provider = require('../lib'); // require('oidc-provider');
+const { Provider } = require('../lib'); // require('oidc-provider');
 const Account = require('../example/support/account');
 const routes = require('../example/routes/koa');
 
--- a/certification/fapi/index.js
+++ b/certification/fapi/index.js
@@ -6,7 +6,7 @@ const jose = require('@panva/jose');
 const helmet = require('koa-helmet');
 const pem = require('https-pem');
 
-const Server = require('../../lib'); // require('oidc-provider');
+const { Provider } = require('../../lib'); // require('oidc-provider');
 
 const OFFICIAL_CERTIFICATION = 'https://www.certification.openid.net';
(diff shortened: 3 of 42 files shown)


alvis:
I can't express how excited I'm seeing this PR. I've been working on a similar definition for oidc-provider in the past couple of days, trying to figure out the types of each function. It's an interesting journey and but I'm more glad to see the official support on it! Thanks!!! 👍

panva:
@alvis cool, you'll be my test subject / guinea pig than :)

alvis:
Just a couple of things to start with

1. There are some missing fields in `OIDCContext`, notably those injected during authorization resume

https://github.com/panva/node-oidc-provider/blob/7071340471601d363c4ad5c75a3c3aa066ee3e2c/lib/actions/authorization/resume.js#L60-L63
https://github.com/panva/node-oidc-provider/blob/7071340471601d363c4ad5c75a3c3aa066ee3e2c/lib/actions/authorization/resume.js#L120

They should be optional fields as they're not always injected

Also, there're couple more.

In my implementation, I have
```ts
  /* ----- variables set in middlewares ----- */

  // see https://github.com/panva/node-oidc-provider/blob/master/lib/actions/authorization/resume.js
  signed?: string[];
  result?: AuthenticationResult;

  // see https://github.com/panva/node-oidc-provider/blob/master/lib/shared/assemble_params.js
  params?: object;

  // see https://github.com/panva/node-oidc-provider/blob/master/lib/shared/selective_body.js
  body?: object;

  // see https://github.com/panva/node-oidc-provider/blob/master/lib/shared/session.js
  session?: Session;

  // ---------------------------------------- //
```

where

```ts
interface AuthenticationResult<
  Meta extends object = { [key: string]: unknown }
> {
  login: Partial<{
    account: string;
    acr: string;
    amr: string[];
    remember: boolean;
    ts: number;
  }>;
  consent: Partial<{
    rejectedScopes: string[];
    rejectedClaims: string[];
  }>;

  meta: Meta;

  [key: string]: object;
}
```

well which is `InteractionResults` in your implementation

2. `Provider` is not exported as default at the moment.
Unless you're going to change it, the correct type sould be `export = Provider`.

Changing to default export is the easiest way to do so, but it'd be a breaking change.
On the other hand, if you opt for the `export = Provider` approach, you'd need to namespace everything under Provider. Ugly. Yes I know...

panva:
ad 1) I believe all are now present
ad 2) I'm not going to change the export to a module with a default, instead I've added `Provider` as an exported property so that you can use that - you can see it in the test file.

alvis:
Sorry for coming back a bit late. There're a couple more issues I found with the definitions.

- I believe more interfaces need to be exported, particularly, `Client`, `Interaction`, `Session`

- `find` is missing in `Interaction`, e.g. the following will result an error

```ts
provider.Interaction.find(id)
```

- `ttlPercentagePassed` in BaseToken should return a number, not boolean
https://github.com/panva/node-oidc-provider/blob/b3d8778985dfd3cf1669d98cb53ed3aec5715ab0/types/index.d.ts#L283

- the signature of Provider.use is incomplete.
https://github.com/panva/node-oidc-provider/blob/b3d8778985dfd3cf1669d98cb53ed3aec5715ab0/types/index.d.ts#L1119

In particular, if one uses koa-router then the following will result in a complaint.
```ts
provider.use(router.routes())
```

A good & proper fix would be using the type from Koa directly, e.g.
```ts
export class Provider extends EventEmitter {
  /* ----- Koa ----- */

  get app(): Koa;
  public env: Koa['env']; // getter and setter
  get callback(): Koa['callback'];
  public keys: Koa['keys']; // getter and setter
  public listen: Koa['listen']; // getter and setter
  public proxy: Koa['proxy']; // getter and setter
  get use(): Koa['use'];
  public subdomainOffset: Koa['subdomainOffset']; // getter and setter

  // ---------------------------------------- //
}
```

- some fields in Client are missing, notably Schema and the constructor. Here is my version.
```ts
interface Schema extends ClientMetadata {}
export class Schema {
  required(): void;
  booleans(): void;
  whens(): void;
  arrays(): void;
  strings(): void;
  normalizeResponseTypes(): void;
  enums(): void;
  webUris(): void;
  scopes(): void;
  postLogoutRedirectUris(): void;
  redirectUris(): void;
  webMessageUris(): void;
  normalizeNativeAppUris(): void;
  checkContacts(): void;
}

export class Client {
  constructor(metadata: ClientMetadata, ctx: Context);

  backchannelLogout(sub: string, sid: string): Promise<void>;

  responseTypeAllowed(type: string): boolean;

  grantTypeAllowed(type: string): boolean;

  redirectUriAllowed(redirectUri: string): boolean;

  checkSessionOriginAllowed(origin: string): boolean;

  webMessageUriAllowed(webMessageUri: string): boolean;

  requestUriAllowed(uri: string): boolean;

  postLogoutRedirectUriAllowed(uri: string): boolean;

  metadata(): ClientMetadata;

  get sectorIdentifier(): string;

  includeSid(): boolean;

  checkClientSecretExpiration(message: string, errorOverride: boolean): void;

  static find<T>(this: { new (): T }, id: string): Promise<T>;

  static needsSecret(metadata: ClientMetadata): boolean;

  static Schema: typeof Schema;
}
```

panva:
Fellas, #569 could use your help.

PR #542, opened 2019-10-03
Update index.d.ts

vanthome:
Add constructor to  OIDCContext class.

For me this is necessary as I want to use it like this:

```
    app.use(async (ctx, next) => {
      if (!ctx.oidc) {
        Object.defineProperty(ctx, 'oidc', { value: new provider.OIDCContext(ctx) });
      }
```

which is impossible with the current typings.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -635,6 +635,7 @@ declare class Client {
 }
 
 declare class OIDCContext {
+  constructor(ctx: Koa.Context);
   readonly route: string;
   uid: string;
   readonly entities: {


panva:
Hi @vanthome, can you elaborate on the use of this a bit more?

vanthome:
Look here: https://github.com/panva/node-oidc-provider/blob/a842b7123854ee5ac5f6b30fd2c53a9ef8b11d5d/lib/helpers/oidc_context.js#L28

The constructor requires a ctx but this is not reflected in the typings.

panva:
I'm aware of the contstructor. I'm just asking if you could elaborate on the use of your own constructed context. I'm curious, never encountered it before.

vanthome:
The code is originally not from me but I think it's inspired by this:
https://github.com/panva/node-oidc-provider/blob/464ebab1d2fbb284f7ce126f43adaa627c31874e/lib/shared/context_ensure_oidc.js

panva:
Oh man, i'm just curious why you need it, that's all :)

vanthome:
Ahh, You mean the module?
We are building a novel e-Commerce platform based on node and uServices and it's the perfect match:
https://github.com/restorecommerce

panva:
When do you need to use this?

```js
    app.use(async (ctx, next) => {
      if (!ctx.oidc) {
        Object.defineProperty(ctx, 'oidc', { value: new provider.OIDCContext(ctx) });
      }
```

vanthome:
we have a setup function that initialized the OIDC provider when the application starts up.

```
      setupOIDC({
        app: this.app,
        issuer,
        logger: this.logger,
        cfg: this.cfg
      });
```

And the previously quoted code is then executed in that `setupOIDC` function.

panva:
I don’t get what that’s good for but anyway, thanks for your feedback and contribution.

vanthome:
great, will show you the whole thing when we published it. Can you then make a new release pls?

panva:
I will do a release when there are other pressing unreleased items or periodically on sunday.

panva:
Released, thanks for your contribution.

---

Please consider [supporting](https://github.com/users/panva/sponsorship) the development of this and my other packages.

PR #557, opened 2019-10-22
Update index.d.ts

perkio:
I needed this inside a custom grant handler that requires access to device code by user:

```
provider.registerGrantType('foo',
  async (ctx, next) => {
    //...
    const deviceCode = await ctx.oidc.provider.DeviceCode.findByUserCode(user_code);
    //...
  });
```

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -432,6 +432,9 @@ declare class DeviceCode extends BaseToken {
     deviceInfo: AnyObject;
     [key: string]: any;
   });
+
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -432,6 +432,9 @@ declare class DeviceCode extends BaseToken {
     deviceInfo: AnyObject;
     [key: string]: any;
   });
+
+  static findByUserCode(userCode: string, options?: { ignoreExpiration?: boolean }): Promise<DeviceCode|undefined>;
```
panva:
```suggestion
  static findByUserCode(userCode: string, options?: { ignoreExpiration?: boolean }): Promise<DeviceCode | undefined>;
```

PR #582, opened 2019-11-15
refactor(typescript): use Koa.ParameterizedContext in typings

ulrikstrid:
I'm using the Koa example to bootstrap a new project and found that this change makes my life a lot easier when using TypeScript and Koa.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -697,9 +697,7 @@ declare class OIDCContext {
   getAccessToken(opts?: { acceptDPoP?: boolean, acceptQueryParam?: boolean }): string;
 }
 
-export interface KoaContextWithOIDC extends Koa.Context {
-  oidc: OIDCContext;
-}
+export type KoaContextWithOIDC = Koa.ParameterizedContext<{oidc: OIDCContext}>;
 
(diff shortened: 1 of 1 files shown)


panva:
@ulrikstrid cool, can you explain how it makes it better? (just wondering)

ulrikstrid:
As I understand it this is the "new" way of using the context. It parameterizes the Koa built-in context instead of extending it which works better with types from the community since the type is the same. I had a bunch of errors when trying to use the current type as is that was solved by this change.

ulrikstrid:
Sidenote: if you're interested in the Koa example working with TypeScript I could extract it before I start making changes to it.

panva:
I would love it if you extended the [ts tests](https://github.com/panva/node-oidc-provider/blob/master/types/oidc-provider-tests.ts) with what you've got and makes sense to have in there.

ulrikstrid:
I extracted it to here: https://github.com/ulrikstrid/node-oidc-provider-typescript-example

I think most of the type errors are because of the Koa stuff that is not present when just using the library as you do in the ts tests.

PR #584, opened 2019-11-15
feat: add script tag nonce resolution helper for session management and wmrm

panva:
@matthew-nichols-westtel

please review, this should help with your CSP nonce setup.


matthew-nichols-westtel:
It looks like scriptNonce will be called and emit a warning if a style-src nonce is set but a script-src nonce is not set

Other than that I don't see anything reading through it

panva:
> It looks like scriptNonce will be called and emit a warning if a style-src nonce is set but a script-src nonce is not set

Is that a big deal? If one's setting CSP nonces it'll set the same one for script and css. I'm willing to look the other way on this one, anything to avoid parsing the csp header.

matthew-nichols-westtel:
As far as I can tell it just means if someone sets a CSS nonce but not a script nonce they'll get a spurious warning, probably not a big deal they can just override scriptNonce to get rid of the warning.

Not a big deal to me. I actually use a different nonce for script and CSS though (don't think it matters here though).

panva:
It doesn’t so long as the function returns the right nonce.

matthew-nichols-westtel:
Fair

PR #586, opened 2019-11-18
[UPDATE] Adding revokeByGrantId into sequelize adapter example

Rogger794:
I add revokeByGrantId into sequelize adapter example and some refactorizations into it like using findByPk instead findByPrimary, also use await instead then. Also throwing error using try catch.

### DIFF
--- a/example/adapters/sequelize.js
+++ b/example/adapters/sequelize.js
@@ -12,13 +12,6 @@ const sequelize = new Sequelize('database', 'username', 'password', {
   storage: 'db.sqlite',
 });
 
-const grantable = new Set([
-  'AccessToken',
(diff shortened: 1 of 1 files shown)


panva:
@Rogger794 feel free to submit a PR to master with the complete adapter again. v5.x PR doesn't matter anymore and the methods aren't used by v5.x anyway...

For v6.x example adapter to be accepted:

1) those try / catches do exactly nothing, they need to be removed.
2) Sequelize.UUIDV4 is not to be used anymore at all, none of the random values set by the provider are uuid
3) session adapter findByUid is missing
4) ... possibly more? Didn't check in depth.

PR #596, opened 2019-11-28
Adding migration using sequelize

Rogger794:
Used for postgresql db


panva:
Hi @Rogger794 

This looks like boilerplate, automatically generated and such, it's missing secondary indexes and doesn't fall into what i'd consider essential for an example adapter implementation. 

Feel free to pull that into your own repo and just update the sequelize adapter with a `@see <url> for ...` js comment instead.

PR #618, opened 2019-12-19
chore(typescript): make claims generic

ulrikstrid:
This should be non-breaking for existing projects but allow to provide claims so that they are infered.

It spread a bit more than I wanted to, let me know if you don't think it's worth it.


panva:
@ulrikstrid please 

- remove the styling changes (indents, double quotes, etc) to leave only the changes in
- add code to https://github.com/panva/node-oidc-provider/blob/master/types/oidc-provider-tests.ts to test the changes

PR #637, opened 2020-01-25
fix: set samesite compatibility cookie as transient when the session is

mikejpeters:
If the session is transient then legacy cookies should also be transient,
so that the session is correctly ended when the browser is closed.

NOTE: I didn't find any tests covering the session middleware, but if you point me in the right direction I can add some


panva:
@mikejpeters I've tested the fix locally, thank you very much for finding and fixing this!

PR #660, opened 2020-02-28
fix(typescript): correct KoaContextWithOIDC definition

### LINKED ISSUE #659: typescript: KoaContextWithOIDC is not correctly defined
In oidc-provider, an OIDCContext is assigned at `ctx.oidc`. However, it's incorrectly defined in the typescript definition due to the type parameter mispositioned in `ParameterizedContext`.

The correct definition of `KoaContextWithOIDC` should be
```ts
export type KoaContextWithOIDC = Koa.ParameterizedContext<{}, {oidc: OIDCContext}>;
```

instead of the one we have today below

https://github.com/panva/node-oidc-provider/blob/71a4a7df23873ef1f6457c3c843c1ff68370bce8/types/index.d.ts#L704

See the definition of `ParameterizedContext`,

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a0d47c0ec44c174267c4f3f85ade76ef91355a62/types/koa/index.d.ts#L704-L706

```ts
type ParameterizedContext<StateT = any, CustomT = {}> = ExtendableContext & {
        state: StateT;
    } & CustomT;
```

Currently, the type definition suggests `ctx.state.oidc` which is wrong.

alvis:
Currently, an OIDCContext is assigned at ctx.oidc. However, it's incorrectly defined in the typescript definition due to the type parameter mispositioned in ParameterizedContext.

fix #659

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -701,7 +701,7 @@ declare class OIDCContext {
   getAccessToken(opts?: { acceptDPoP?: boolean, acceptQueryParam?: boolean }): string;
 }
 
-export type KoaContextWithOIDC = Koa.ParameterizedContext<{oidc: OIDCContext}>;
+export type KoaContextWithOIDC = Koa.ParameterizedContext<{}, { oidc: OIDCContext }>;
 
 export const DYNAMIC_SCOPE_LABEL: symbol;
 


panva:
Thank you very much @alvis

PR #662, opened 2020-02-28
fix(typescript): correct KoaContextWithOIDC definition, take 2

alvis:
Sorry, my bad. I missed the default state and context in KoaContextWithOIDC, making it too restrictive to be extended.


panva:
@alvis can you check the diff to make sure this is what you want to push?

alvis:
Fixed. Forgot to pull before merging 😓

PR #665, opened 2020-03-03
fix(typescript): adapter methods typings and interface name typo

fiznool:
The example provided in the docs shows the [sequelize adapter returning a `consumed` property if the token has been consumed](https://github.com/panva/node-oidc-provider/blob/30e439ac525d6d10361537046ef9682d736284ea/example/adapters/sequelize.js#L69-L94), but this property is not present in the typings. This causes a TypeScript error with the latest version of TypeScript (3.8.3).

The fix here is to return a new interface from the `find*` methods of the adapter class which includes this `consumed` property, along with all of the other possible properties from`AdapterPayload`.

Additionally, a devDependency on TypeScript was added to satisy a warning when running `npm install` (`dtslint` has a peerDependency on typescript).

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,8 @@
     "paseto": "^1.0.5",
     "sinon": "^9.0.0",
     "supertest": "^4.0.2",
-    "timekeeper": "^2.2.0"
+    "timekeeper": "^2.2.0",
+    "typescript": "^3.8.3"
   },
   "engines": {
     "node": "^10.13.0 || >=12.0.0"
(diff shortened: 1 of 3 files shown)


panva:
@fiznool the consumed property can actually be anything truthy/falsy, literally `any` type, so let's do that.

Rather than adding a new interface just add `consumed?: any` to AdapterPayload.

fiznool:
I did wonder about adding it to the `AdapterPayload`, but I wasn't sure if it was correct to have a `consumed` property in the data sent into the `upsert` method? This is why I created a new interface.

panva:
> I did wonder about adding it to the `AdapterPayload`, but I wasn't sure if it was correct to have a `consumed` property in the data sent into the `upsert` method? This is why I created a new interface.

I get it and I'm fine with it.

PR #716, opened 2020-05-11
refactor: removed inline script from form post html

rubenvanrooij:
Helps in creating a more secure CSP.

### DIFF
--- a/test/device_code/code_verification_endpoint.test.js
+++ b/test/device_code/code_verification_endpoint.test.js
@@ -31,7 +31,8 @@ describe('GET code_verification endpoint', () => {
         .query({ user_code: '123-456-789' })
         .expect(200)
         .expect('content-type', 'text/html; charset=utf-8')
-        .expect(/<body onload="javascript:document\.forms\[0]\.submit\(\)"/)
+        .expect(/<body>/)
+        .expect(/document.addEventListener\('DOMContentLoaded', function \(\) { document.forms\[0\].submit\(\) }\);/)
         .expect(({ text }) => {
           ({ state: { secret } } = this.getSession());
           expect(text).to.match(new RegExp(`input type="hidden" name="xsrf" value="${secret}"`));
--- a/test/device_code/code_verification_endpoint.test.js
+++ b/test/device_code/code_verification_endpoint.test.js
@@ -31,7 +31,6 @@ describe('GET code_verification endpoint', () => {
         .query({ user_code: '123-456-789' })
         .expect(200)
         .expect('content-type', 'text/html; charset=utf-8')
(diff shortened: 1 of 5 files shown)


panva:
> Helps in creating a more secure CSP.

@rubenvanrooij can you elaborate?

rubenvanrooij:
> > Helps in creating a more secure CSP.
> 
> @rubenvanrooij can you elaborate?

Sure! You cannot use a SHA-256 hash to whitelist an inline event handler in CSP2.0. This is because the context of the javascript cannot be determined. E.g. x.deletAll() is ok on a big red button after several confirmation steps. But is not ok when injected on the body element. 
Therefore the JS need to be in it's own <script> tag so that you're forced to reference the exact element(s) you want to have the (in this case) eventlisteners attached to. 

In CSP you could enable `unsafe-inline` or `unsafe-hashes` but imho this wouldn't be preferable. So with the JS moved to its own <script> tag we can easily add the following hash: 

`sha256-RdUNylKSz5sOn0fkU1eqAq2AfjeUgGQlHsNg7gbsfrU=

to our CSP configuration to allow the JS to run.

test/device_code/code_verification_endpoint.test.js
```diff
@@ -31,7 +31,8 @@ describe('GET code_verification endpoint', () => {
         .query({ user_code: '123-456-789' })
         .expect(200)
         .expect('content-type', 'text/html; charset=utf-8')
-        .expect(/<body onload="javascript:document\.forms\[0]\.submit\(\)"/)
+        .expect(/<body>/)
```
panva, line 34: lets remove this expect

test/device_code/device_resume.test.js
```diff
@@ -392,7 +392,7 @@ describe('device interaction resume /device/:user_code/:uid/', () => {
         await this.agent.get(path)
           .expect(200)
           .expect('content-type', 'text/html; charset=utf-8')
-          .expect(/<body onload="javascript:document\.forms\[0]\.submit\(\)"/)
+          .expect(/<body>/)
```
panva, line 395: remove this line and add the one asserting that a self submitting code is present.

test/interaction/interaction.test.js
```diff
@@ -394,7 +394,7 @@ describe('resume after interaction', () => {
       await this.agent.get('/auth/resume')
         .expect(200)
         .expect('content-type', 'text/html; charset=utf-8')
-        .expect(/<body onload="javascript:document\.forms\[0]\.submit\(\)"/)
+        .expect(/<body>/)
```
panva, line 397: remove this line and add the one asserting that a self submitting code is present.

PR #718, opened 2020-05-14
fix(typescript): undefined return for DefaultPolicy.get and Checks.get

adriencohen:


panva:
@adriencohen thank you

PR #732, opened 2020-06-13
type fix

qwertypnk:
hello, isn't it supposed to be without typeof?


panva:
No, it is supposed to be typeof.

qwertypnk:
actually you're right, sorry

PR #735, opened 2020-06-15
fix: correct jwtIntrospection draft ack version

### LINKED ISSUE #734: Unable to ack latest jwtIntrospection draft version due too TypeScript typings
**Describe the bug**
The latest draft version for the jwtintrospection cannot be acknowledged due to the TypeScript typings.

**To Reproduce**
Provider and Client configuration: (inline or gist)

To crash on runtime with the error `An unacknowledged version of a draft feature is included in this oidc-provider version.`
```js
const provider = {
  ...,
  features: {
    jwtIntrospection: {
      enabled: true,
      ack: 'draft-09' as const
    },
  }
}
```

To crash on build time with the error `Type '"draft-08"' is not assignable to type '"draft-09" | undefined'`.
```js
const provider = {
  ...,
  features: {
    jwtIntrospection: {
      enabled: true,
      ack: 'draft-08' as const
    },
  }
}
```

Steps to reproduce the behaviour:

1.  Enable the `jwtIntrospection` feature and ack as either of the above supplied configurations
2.  Compile and/or run the application

**Expected behaviour**
The latest draft-version implemented by this library (which I assume to be `draft-09`) should be acknowledgeable following the TypeScript definitions using the following configuration:

```js
const provider = {
  ...,
  features: {
    jwtIntrospection: {
      enabled: true,
      ack: 'draft-09' as const
    },
  }
}
```

**Environment:**
 - oidc-provider version: 6.27.0
 - node version: 14.3.0

**Additional context**
It seems that this problem was "introduced" in this commit 219e8c3df230737c175e9fb9b318c4bd1c0ef011, the typings and functionality are updated in this commit but the acknowledged drafts are NOT updated. It further seems that this introduced a breaking change even if the user of this library acknowledged the previously correct draft version (draft-08).

 - [x] the bug is happening on latest oidc-provider too.
 - [x] i have tried DEBUG (see README) and can see the issue is with the provider and not my code.
 - [x] i have searched the issues tracker on github for similar issues and couldn't find anything related.
 - [x] i have searched the [FAQ](https://github.com/panva/node-oidc-provider/blob/master/docs/README.md#faq) and couldn't find anything related.

**Logs**

```
TSError: ⨯ Unable to compile TypeScript:
src/authentication/config.ts:223:53 - error TS2345: Argument of type '{ adapter: any; jwks: any; claims: { [key: string]: string[]; }; clients: never[]; extraClientMetadata: { properties: string[]; validator: (key: string, value: any) => void; }; features: { claimsParameter: { enabled: boolean; }; ... 6 more ...; frontchannelLogout: { ...; }; }; ... 8 more ...; extraAccessTokenClaims(...' is not assignable to parameter of type 'Configuration'.
  The types of 'features.jwtIntrospection.ack' are incompatible between these types.
    Type '"draft-08"' is not assignable to type '"draft-09" | undefined'.
```

```
NOTICE: The following draft features are enabled and their implemented version not acknowledged
NOTICE:   - JWT Response for OAuth Token Introspection - draft 08 (This is an IETF OAuth Working Group draft. URL: https://tools.ietf.org/html/draft-ietf-oauth-jwt-introspection-response-08. Acknowledging this feature's implemented version can be done with the string 'draft-08')
NOTICE: Breaking changes between draft version updates may occur and these will be published as MINOR semver oidc-provider updates.
NOTICE: You may disable this notice and these potentially breaking updates by acknowledging the current draft version. See https://github.com/panva/node-oidc-provider/tree/v6.27.0/docs/README.md#features

$PATH/node_modules/oidc-provider/lib/helpers/configuration.js:539
        throw new TypeError('An unacknowledged version of a draft feature is included in this oidc-provider version.');
```

milo526:
Adds the expected acknowledgement for `draft-09` for the `jwtIntrospection`.

This enables users of this library to acknowledge the proper jwt introspection draft version that this library is currently using, it further prevents users who have acknowledged `draft-08` previously from unexpectedly breaking their application when updating the library.

This MR fixes #734

### DIFF
--- a/lib/helpers/features.js
+++ b/lib/helpers/features.js
@@ -46,10 +46,10 @@ const DRAFTS = new Map(Object.entries({
     version: [2, 'draft-02'],
   },
   jwtIntrospection: {
-    name: 'JWT Response for OAuth Token Introspection - draft 08',
+    name: 'JWT Response for OAuth Token Introspection - draft 09',
     type: 'IETF OAuth Working Group draft',
-    url: 'https://tools.ietf.org/html/draft-ietf-oauth-jwt-introspection-response-08',
-    version: [8, 'draft-08'],
+    url: 'https://tools.ietf.org/html/draft-ietf-oauth-jwt-introspection-response-09',
+    version: ['draft-09'],
   },
   jwtResponseModes: {
     name: 'JWT Secured Authorization Response Mode for OAuth 2.0 - draft 02',


lib/helpers/features.js
```diff
@@ -46,10 +46,10 @@ const DRAFTS = new Map(Object.entries({
     version: [2, 'draft-02'],
   },
   jwtIntrospection: {
-    name: 'JWT Response for OAuth Token Introspection - draft 08',
+    name: 'JWT Response for OAuth Token Introspection - draft 09',
     type: 'IETF OAuth Working Group draft',
-    url: 'https://tools.ietf.org/html/draft-ietf-oauth-jwt-introspection-response-08',
-    version: [8, 'draft-08'],
+    url: 'https://tools.ietf.org/html/draft-ietf-oauth-jwt-introspection-response-09',
+    version: [9, 'draft-09'],
```
panva, line 52: Remove the 9 please, just leave the string in.

milo526: Resolved as amendment to the original commit.

PR #748, opened 2020-07-08
Remove discovery

carolineBda:
Remove discovery option as it throws error "features are no longer enabled/disabled with a boolean value, please see the docs"


panva:
I'm closing this. A PR open against the v5.x branch when the feature options were simple booleans. It's aligned. In v6.x as well.

PR #776, opened 2020-08-24
feat(typescript): allows adapter factory in typings

twawszczak:
Change typings to allow for the functional adapter creator.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -814,6 +814,8 @@ export interface Adapter {
   revokeByGrantId(grantId: string): Promise<undefined | void>;
 }
 
+export type AdapterFactory = (name: string) => Adapter;
+
 export interface AdapterConstructor {
   new(name: string): Adapter;
 }
@@ -832,7 +834,7 @@ export interface CookiesSetOptions {
 export interface Configuration {
   acrValues?: string[] | Set<string>;
 
-  adapter?: AdapterConstructor;
+  adapter?: AdapterConstructor | AdapterFactory;
 
   claims?: {
     [key: string]: null | string[]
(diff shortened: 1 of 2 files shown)


panva:
@twawszczak thank you, the fix is now released on npm as part of v6.29.3

jstewmon:
👋 Hi, I've been looking over how the `adapter` configuration property is used, and wanted to note that this type definition is overly permissive, and the supporting test would result in a runtime error.

The `initializeAdapter` function requires that the `adapter` have a `prototype` property with `constructor.name`, which the following will not have:
- arrow functions
- shorthand methods
- class methods

I say "somewhat ambiguous" because AFAIK TypeScript will not consider a function type to be compatible with an interface which defines a construct signature.

These two ways of providing the adapter will work at runtime:

```js
new Provider(issuer, { adapter: function() { return {}; } });
new Provider(issuer, { adapter: class { } })
```

But these will all fail at runtime despite satisfying the `AdapterFactory` interface if the return type satisfies the `Adapter` interface:

```js
new Provider(issuer, { adapter() { return {} } });
new Provider(issuer, { adapter: () => { return {} } });
// Error: Expected "adapter" to be a constructor, provide a valid adapter in Provider config.
```

Since the `adapter` must be newable, I think it would be preferable to remove the `AdapaterFactory`, so that these cases which are known to fail will not pass the type checker.

For TypeScript users who wish to provide a plain function, they can use a type assertion:

```ts
new Provider("https://op.example.com", {
  adapter: function (name: string): Adapter & {name: string} {
    return {
      name,
      async upsert(id: string, payload: AdapterPayload, expiresIn: number) {},
      async consume(id: string) {},
      async destroy(id: string) {},
      async revokeByGrantId(grantId: string) {},
      async find(id: string) {
        return {
          client_id: "...",
        };
      },
      async findByUserCode(userCode: string) {},
      async findByUid(uid: string) {},
    };
  } as unknown as AdapterConstructor,
});
```

PR #786, opened 2020-09-18
fix(typescript): findAccount may return undefined

jstewmon:
All findAccount call sites handle the possibility that a falsy value is
returned, indicating that no account was found.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -14,7 +14,7 @@ export {};
 
 export type CanBePromise<T> = Promise<T> | T;
 export type RetryFunction = (retry: number, error: Error) => number;
-export type FindAccount = (ctx: KoaContextWithOIDC, sub: string, token?: AuthorizationCode | AccessToken | DeviceCode) => CanBePromise<Account>;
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -14,7 +14,7 @@ export {};
 
 export type CanBePromise<T> = Promise<T> | T;
 export type RetryFunction = (retry: number, error: Error) => number;
-export type FindAccount = (ctx: KoaContextWithOIDC, sub: string, token?: AuthorizationCode | AccessToken | DeviceCode) => CanBePromise<Account>;
+export type FindAccount = (ctx: KoaContextWithOIDC, sub: string, token?: AuthorizationCode | AccessToken | DeviceCode) => CanBePromise<Account | undefined | null>;
```
panva, line 17:
```suggestion
export type FindAccount = (ctx: KoaContextWithOIDC, sub: string, token?: AuthorizationCode | AccessToken | DeviceCode) => CanBePromise<Account | undefined>;
```

PR #794, opened 2020-10-03
fix: loopback redirectUriAllowed protocol check

mureinik:
The condition `!parsed.protocol === 'http:'` always returns `false`. The `!` operator has a higher precedence than the `===` operator. `parsed.protocol` is a string, so negating it would be `false`. Then, it's clear that `false === 'http:'` is always `false`.

It seems that the intended check was `parsed.protocol !== 'http:'`.


panva (approved):
Thanks for catching this, clearly an oversight on my part when this was being refactored

PR #805, opened 2020-10-25
chore: upgrade the `got`-dependency

weyert:
Upgraded the `got`-dependency to the same version that is being used by sibling package `node-openid-client`


panva:
9.6.0 is locked in for v6.x. Can't do much about it, upgrading got would be a breaking change. v7.x coming eventually is going to sport the latest got.

weyert:
Oh okay, then I will keep using v3.x of `openid-client` until v7 :)

panva:
I don't get why. The modules are independent.

weyert:
Yeah, I understand, I will raise a ticket in `openid-client` about it

panva:
If its about conflicting types from @types and the bundled one from updated got, that's a TS deficiency i cannot solve for you.

PR #821, opened 2020-11-17
export AccessToken + AuthorizationCode + DeviceCode

brandonros:
it's impossible to define a good `expiresWithSession` configuration override without these types exported

```
const expiresWithSession = async (ctx: KoaContextWithOIDC, token: oidc.AccessToken | oidc.AuthorizationCode | oidc.DeviceCode) => {
  return false
}
```


panva:
`type AccessToken = InstanceType<Provider['AccessToken']>` not good enough? I am not okay exposing the class like so because it is not actually what the API is.

is `export type ...` an option?

PR #833, opened 2020-12-05
fix(typescript): interaction result may be undefined

adriencohen:
Fix result typing which can be undefined when getting the interaction from `interactionDetails()`


panva:
@adriencohen thank you!

PR #838, opened 2020-12-12
fix(typescript): Use interface instead class and export it to allow usage in typescript project.

Romakita:
Hello Team,
To begin, thanks for your amazing work about this node module. 

## Description

I make this PR because I want to integrate node-oidc-provider in the [Ts.ED framework](https://tsed.io). All works fine excepted when I compile the typescript module because some of Provider methods returns unnamed classes (like Interaction). The current typing file implementation is aligned with the node-oidc-provider structure but because `Interaction` is a declared class instead of exported class, isn't possible to reuse the type correctly.

Example:
```typescript
declare class Interaction extends BaseModel
```

To explain my problem, here a small code example with the Ts.ED framework:
```typescript
@Injectable()
@Scope(ProviderType.REQUEST)
export class InteractionService {
  @Inject()
  provider: Provider;

  @Inject()
  context: PlatformContext;

  interactionDetails() { // Error when compiling with TS
    return this.provider.interactionDetails(
       this.context.getReq(),
       this.context.getRes()
    )
  }
}
```

The resulting error is the following:

```
Return type of public method from exported class has or is using name 'Interaction' from external module ".../node_modules/oidc-provider/types/index" but cannot be named.
```

## Solution

To stay aligned with the node-oidc-provider we cannot change the `declare` keyword by `export` because, the module doesn't export directly the Interaction class. The solution is to transform class to interface.

Here the change:

```diff
+ export interface InteractionConstructor extends BaseModelConstructor {
+  new(): Interaction;
+ }

+ export interface Interaction extends BaseModel {
- declare class Interaction extends BaseModel {
```
This solution is totally backward compatible :)

I hope this PR will help any other typescript project in the same situation ;)

See you
Romain


panva:
@Romakita thank you for the PR, it'll take me a bit to review but it seems to be just the right thing!

Romakita:
@panva Thanks for your time ;)

types/index.d.ts
```diff
@@ -1490,23 +1562,23 @@ export class Provider extends events.EventEmitter {
...
-  readonly AccessToken: typeof AccessToken;
-  readonly InitialAccessToken: typeof InitialAccessToken;
-  readonly RefreshToken: typeof RefreshToken;
-  readonly AuthorizationCode: typeof AuthorizationCode;
-  readonly RegistrationAccessToken: typeof RegistrationAccessToken;
-  readonly PushedAuthorizationRequest: typeof PushedAuthorizationRequest;
-  readonly ClientCredentials: typeof ClientCredentials;
-  readonly DeviceCode: typeof DeviceCode;
-  readonly BaseToken: typeof BaseToken;
+  readonly Client: ClientConstructor;
```
panva, line 1565: Correct me if i'm wrong but doesn't this change mean that all of [this](https://github.com/panva/node-oidc-provider/pull/838/files?diff=split&w=1#diff-093ad82a25aee498b11febf1cdcb6546e4d223ffcb49ed69cc275ac27ce0ccceR642-R652) and other potentially useful prototype overloadable methods won't be overloadable in TS when accessed through e.g. `provider.Client.prototype.redirectUriAllowed`?

dtto for essentially all the models.

What everyone can do now to get the types, altho not exported, if they need them is, e.g.

```
type Client = InstanceType<Provider['Client']>;
```

Romakita: I am not sure I fully understand. But if you talk about the possibility of doing the inheritance like following or override method, it's always possible:

Capture d’écran 2020-12-14 à 20 57 14

Unfortunately, I tried to use your example, but it doesn't work:

```typescript
type Client = InstanceType<Provider['Client']>;
``` 
Typescript considers the instanceType as any. In the capture, we can see that the fields aren't available:

Capture d’écran 2020-12-14 à 20 41 00

panva: I don't care about extending classes, only important thing to me is overloading the prototype methods i consider part of the semver contract. With `provider.Client` being typed as ClientConstructor that only has `find` as a static method all the actual Client instance methods are not typed now.

See the before/after comparison of master/your PR

Screenshot 2020-12-14 at 21 06 05
Screenshot 2020-12-14 at 21 06 21

panva: And the InstanceType method works just fine.

Screenshot 2020-12-14 at 21 10 25

proof: without casting the result as Client.

Screenshot 2020-12-14 at 21 10 35

Romakita: Ok my bad, I realize the problem is on my side. My apologize for the wasted time :) and thanks for Tips ;)
I close the PR now and thanks again.

panva: You're welcome and thank you! 

Please consider [supporting](https://github.com/panva/node-oidc-provider?sponsor=1) the library if it provides value to you or your company and this was of help to you. Supporting the library means, amongst other things, that such support will be available in the future.

Romakita: Yes sure, I’m the author of Ts.ED framework and I understand the importance of the support ;). When the new module will be ready, I’ll publish a page to present the oidc-provider and the integration with the framework! I hope the module will help our respective communities ;)

See you

PR #854, opened 2021-01-11
fix: missing login prompt details

### LINKED ISSUE #853: Login interaction details
https://github.com/panva/node-oidc-provider/blob/2f9def003ee3194588f98947abeadea12996ef76/lib/helpers/interaction_policy/prompts/login/index.js#L18-L22

Hello,
I think there is an error in the order of the ternary expressions : in the final object the value is set when it is undefined and omitted when it is defined.

HNicolas:
Hello,
This should fix #853 if it is an error.


panva (approved):

panva:
@HNicolas thank you

PR #860, opened 2021-01-18
docs: update interactions.policy default value

HNicolas:
I updated the doc in order correct the default value for `interactions.policy`.


panva (changes requested):
Please make the change in checks.txt and then re-run `node ./docs/update-configuration.js`. Commit both file changes.

PR #882, opened 2021-02-01
Provider configuration getter

twawszczak-wr:
This PR proposal contains a lot of changed files, but most of them are just alignments of the existing code and could be easily reverted and excluded from the PR. There is only one important part in the `lib/provider.js` file:

```js
configuration(path) {
  return instance(this).configuration(path);
}
```
It exposes simplified interface for a configuration read. Why? It allows to simplify code, resign from `helpers/weak_cache` dependency, what is especially helpful for for the TypeScript users and the external lib extensions. (OIDC context is frequently passed to configuration callbacks)

@panva What do you think? :)

### DIFF
--- a/lib/actions/authorization/assign_claims.js
+++ b/lib/actions/authorization/assign_claims.js
@@ -1,5 +1,4 @@
 const merge = require('../../helpers/_/merge');
-const instance = require('../../helpers/weak_cache');
 
 /*
  * Merges requested claims with auth_time as requested if max_age is provided or require_auth_time
@@ -10,7 +9,7 @@ const instance = require('../../helpers/weak_cache');
 module.exports = function assignClaims(ctx, next) {
   const { params } = ctx.oidc;
 
-  if (params.claims !== undefined && instance(ctx.oidc.provider).configuration('features.claimsParameter.enabled')) {
+  if (params.claims !== undefined && ctx.oidc.provider.configuration('features.claimsParameter.enabled')) {
     ctx.oidc.claims = JSON.parse(params.claims);
   }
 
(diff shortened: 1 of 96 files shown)


panva:
The configuration/weak_cache helper is omitted from a public API on purpose. Until this moment absolutely every use-case i've came across can be solved without it.

panva:
> OIDC context is frequently passed to configuration callbacks

It is also available as `provider.OIDCContext` getter.

Again, up until this point everything that was thrown at this issue tracker regarding this subject could be solved without this helper. Making it part of the public API puts too much constrain on my ability to restructure things as i see fit to solve other problems.

twawszczak-wr:
@panva I appreciate your concepts. (and also high efficiency in the stuff rejecting :)) I would like only to ask, if you could reveal it, what is the purpose of excluding configuration reference from the public API? 

For example my use case - I want to use global configuration `clockTolerance` property for the `software_statement` JWT parser, similar to this one: https://github.com/panva/node-oidc-provider/blob/main/docs/README.md#extraclientmetadatavalidator It seems reasonably to fetch this property from `oidcContext` passed to `extraClientMetadata.validator` callback. (ofc a workaround could be used, but why?)

Another examples - custom grants implementations. https://github.com/panva/node-oidc-provider/blob/main/docs/README.md#custom-grant-types Hard to implement without an access to the global configuration, You are redirecting to standard grants implementations, and there you are using `weak_cache` wrapper everywhere. 

Anyway, I appreciate your hard work and great libraries. Regards.

panva:
@twawszczak-wr 

ad clockTolerance configuration) since you can explicitly configure this property i would assume it's just a matter of keeping a reference to your explicit configuration to get this value.

ad custom grants) pretty much the same, it's mostly used to look up feature configuration that you know how to reach since it's part of your explicit provider configuration.

PR #913, opened 2021-02-24
Support passing a function for cookies names configuration

JonnyBGod:
This is a WIP, it is missing tests.
But meanwhile it would be good to have your opinion on the current work.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -2131,6 +2131,15 @@ _**default value**_:
 }
 ```
 
+Configure `cookies.names` like so:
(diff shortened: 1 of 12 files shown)


panva:
> But meanwhile it would be good to have your opinion on the current work.

Without the details on how and for what purpose might one use it, plus the sure to follow debate, I can only say this work is a bit premature ;) I'm going to close for now and we can have that discussion either in your original discussion thread, or preferrably, a new one in the Ideas section.

PR #924, opened 2021-03-02
fix(typescript): Interaction.prototype.session structure

jstewmon:
Interaction.session is always a plain object - its type was changed to
a union type including Session when the Interaction constructor was
updated to handle payload.session being an instance of Session, but in
such a case, the Session object is transformed to a plain object.

The jti property appears to have been an extraneous duplicate of the
cookie property - a Session object's jti property is mapped to the
cookie property of the plain object.

The uid property was simply missing from the type definition.


panva:
@jstewmon thank you!

PR #954, opened 2021-03-22
docs: add new recipes "Avoid excuting expensive operations in findAccount"

qfdk:
Hello,

I have created a new recipes for this project :)

Have a nice day.

### DIFF
--- a/lib/helpers/jwt.js
+++ b/lib/helpers/jwt.js
@@ -183,6 +183,11 @@ class JWT {
       alg, enc, ...fields,
     };
 
+    if (key.type === 'private') {
+      // eslint-disable-next-line no-param-reassign
+      key = createPublicKey(key);
+    }
+
     return new CompactEncrypt(Buffer.from(cleartext))
       .setProtectedHeader(header)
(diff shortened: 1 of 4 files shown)


panva:
Hi @qfdk, thanks for following up on this.


Is there really a need for a recipe that just says "you can memoize and cache results of helper functions"?

qfdk:
Hello @panva ,

Thanks for your quickly response. 

In my opinion, it’s useful to mention in documentation. I don’t know if it’s a good idea to put thisin recipes, maybe `add a pointer to documentation `you can memoize or cache results of helper functions` with these samples in findAccount part ?

It shows also when findAccound called.

In my case, I took time to find out  `memoize or cache results of helper functions`  in the documentation and find some similar discussion. To show some sample code is a good way to help others.

panva:
My point is that one shouldn't need to be told by a documentation that when their particular implementation is heavy on the processing side they should optimize to not have it called that often...

qfdk:
Thanks for your response. I just want to share the idea with others :)

PR #973, opened 2021-04-09
fix: Add Node.js v15 support

Romakita:
Hello @panva 

This PR is to fix the issue over the following error:
```
error oidc-provider@7.3.0: The engine "node" is incompatible with this module. Expected version "^12.19.0 || ^14.15.0". Got "15.14.0"
```

See you
Romain


panva:
With v7.x the Node.js runtime version policy changed. Version 12.19.0 is now the minimum required runtime and ONLY LTS releases are supported. This means "Current" Node releases are not officially supported and you may get mixed results when using them.

If you want to use those versions that's fine and i don't expect issues. The next release line to be supported is v16.x once it enters and LTS stage on 2021-10-26.

Until then "^12.19.0 || ^14.15.0" is the right version to check for, as denoted by the package.json `engines` entry as well.

Romakita:
Hello @panva,

You have to stop being so closed ^^. But I understand your point.
  
In my case, I don't want to develop an OIDC server on this version, but just to avoid having an error related to the installation of your framework on my framework with the CI (see here the problem: https://github.com/TypedProject/tsed/runs/2305449726?check_suite_focus=true)

I take the time to provide a module that quickly integrates with Ts.ED! I lost my morning to migrate the @tsed/oidc-provider (https://tsed.io/tutorials/oidc.html) on your v7 without a clear migration guide and now I have an issue because you've set the node engine XD. I'll use `--ignore-engines` to fix that ;)

See you
Romain

thecodejack:
I don't understand blocking other node engines. Current or Non LTS releases are as stable as LTS. And unless a node module is failing, providing engines config may not help as well.

panva:
> I don't understand blocking other node engines. Current or Non LTS releases are as stable as LTS. And unless a node module is failing, providing engines config may not help as well.

It's not blocking anything. It sets an expectation as to what support could be given for. You're free to use whatever engine you want but you'll be on your own wrt e.g. http module breaking, crypto breaking, and so on. Both of which btw happened with 15.x and 14.x before it entered lts.

panva:
> Current or Non LTS releases are as stable as LTS

Also no, they're not.

thecodejack:
Problem for us is, node engines gets limited to whatever is been mentioned. We have to add `--ignore-engines` to avoid such issues.

thecodejack:
>Also no, they're not.

They are. IMO LTS versions are maintained longer means gets longer updates(mostly fixes) while others won't have longer updates but expect new features available.

panva:
> > Also no, they're not.
> 
> They are.

🤷

PR #1024, opened 2021-06-12
chore: node engines support update in package.json

thecodejack:
Tested with node v15.9.0 and v16.3.0


panva:
Hi. This will not happen until v16.x enters LTS. More in https://github.com/panva/node-oidc-provider/pull/973#issuecomment-816649081

PR #1036, opened 2021-06-24
loadExistingGrant - defaults implementation fix

MichaelRo:
Just a small fix, in case session hasn't been created


panva (changes requested):
Hi @MichaelRo, thanks for submitting a PR. Couple notes:

- we can't use this syntax on node 12
- can you expand on the why? maybe add a test case or concrete steps to reproduce an error

panva:
happy to reopen when you have the time to provide the details

PR #1038, opened 2021-06-24
fix: correctly check access token audience when issuing ID tokens

roc13x:
Fixes an issue where ID tokens would not include account claims, even if the access token was not usable for the userInfo endpoint (contained an aud value).

Currently, if you use the resourceIndicators feature and issue an access token in JWT format, then the returned ID token will not contain any account claims. This is despite the fact that the JWT access token is not usable on the userInfo endpoint.

Problem is caused by the code checking the access token for an `aud` value. It's checking the generated token string rather then the original token object.


panva:
@roc13x nice catch, thank you for bringing it up.

PR #1052, opened 2021-07-15
MR to resolve script error from form_post.js

vaibhavmedhekar92:
Background to this MR: 
When a web browser is embedded as control, it runs in IE7 document mode by default even though the latest version of IE is installed on machine. Because of this WPF desktop application is throwing script error to execute "addEventListener" function from form_post.js

Resolution:
The solution to this issue is to add below meta tag inside head tag at form_post.js file. This will tell browser to use latest version of IE installed on machine to run the document.
<meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">

It would be very helpful it this change is available in 6.31.0 or may be newly published version like 6.31.0-a as we are using 6.31.0 version in our services.

### DIFF
--- a/lib/response_modes/form_post.js
+++ b/lib/response_modes/form_post.js
@@ -15,6 +15,7 @@ module.exports = function formPost(ctx, action, inputs) {
 
   ctx.body = `<!DOCTYPE html>
 <head>
+  <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
   <title>Submitting Callback</title>
   <script>
     document.addEventListener('DOMContentLoaded', function () { document.forms[0].submit() });


panva:
I'm afraid i'll need a deeper explanation. "When a web browser is embedded" itself tells me adding this makes it easier to embed the auth flows outside of an actual browser? If so, I'm not sure it aligns with the best practice.

vaibhavmedhekar92:
Hi @panva 

Our product bucket have few WPF based windows applications which uses OAuth 2.0 for user authentication.
While logging in to our windows application, login window is opened as a web browser control. Here user can enter his credentials and login to our app.
In few scenarios, we have observed that we are getting script error on login page stating: 
**Object doesn't support property or method "addEventListener"**. The screenshot for the same is attached here.
login_error

Upon further analysis we have found that in our recent releases we have bumped the version of oidc-provider from v6.23.2 to v6.31.0
In v6.31.0 we have observed below addition in the form_post.js file:

<script>
    document.addEventListener('DOMContentLoaded', function () { document.forms[0].submit() });
</script>

The commit in oidc-provider responsible for this change is:
https://github.com/panva/node-oidc-provider/commit/9270b613b391b0805aa625bf2a26ab3aaa128d8f
  
While researching we have found the similar issue reported due to above change:
https://github.com/IdentityServer/IdentityServer4/issues/3128

Upon analysis we have found that when web browser is loaded as control (in our case for login page), it always run in IE7 document mode by default.
Because of this behavior we are seeing script error in addEventListener.
The fix for this issue is to tell browser to load the document in IE version installed on the user's machine.
Adding below tag in the head section of html document in form_post.js page serves this fix.

<meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">

We have verified the fix in our development environment.
As we are using v6.31.0 of oidc-provider in production, we request you to provide this fix in the same version.

Thanks & Regards,
Vaibhav Medhekar

panva:
I have just now noticed that the PR is targeting v6.x. I'm going to cherry-pick it to 'main' to apply to the latest release. If you wish it applied to a legacy release line please consider sponsoring the library as I only consider releasing for those lines for supporters that have been with the project for a bit, unless it's for a security fix.

vaibhavmedhekar92:
Hi @panva 
I don't see the script error fix pushed to the main branch yet. Could you please check and let me know.
Thanks & Regards,
Vaibhav Medhekar

panva:
@vaibhavmedhekar92 https://github.com/panva/node-oidc-provider/releases/tag/v7.5.4

Issue #1047, opened 2021-07-14
Different Client ID, same browser session, it does not ask to authenticate  


cloudstorydev:
**Describe the bug**
Open a browser, authenticate using client_id="xyz", and success
Same browser session, authenticate using client_id = "abc", redirect back to app without login prompt. It apparently picked up the same session from the database with client_id xyz, based on the logs.  Token issued are correct (abc), but why no login prompt?

 
**Environment:**
 - oidc-provider version:   7.1.1
 - node version:  v12.22.1
 
**Logs**
 After trying to login using second client_id (abc):

Executing (default): SELECT "id", "uid", "data", "expiresAt", "consumedAt", "createdAt", "updatedAt" FROM "idp_sessions" AS "Session" WHERE "Session"."id" = 'Uog8Kd0zeYvoHRroqGbz1';
Executing (default): SELECT "id", "data", "expiresAt", "consumedAt", "createdAt", "updatedAt" FROM "idp_clients" AS "Client" WHERE "Client"."id" = 'abc';
resource indicator:  http:/localhost:8080 Client {
  applicationType: 'web',
  grantTypes: [ 'authorization_code', 'refresh_token' ],
  idTokenSignedResponseAlg: 'RS256',
  postLogoutRedirectUris: [
    'https://abc.idm.test/oidc/logout',
    'https://abc.lms.test/oidc/logout',
    'http://localhost:8000/oidc/logout'
  ],
  requireAuthTime: false,
  responseTypes: [ 'code' ],
  subjectType: 'public',
  tokenEndpointAuthMethod: 'none',
  introspectionEndpointAuthMethod: 'none',
  requireSignedRequestObject: false,
  requestUris: [],
  clientId: 'abc',
  clientSecret: 'secret',
  redirectUris: [
    'https://abc.idm.test/oidc/callback',
    'https://abc.lms.test/oidc/callback',
    'http://localhost:8000/oidc/callback'
  ]
}

====== 

Executing (default): UPDATE "idp_authorization_codes" SET "consumedAt"=$1,"updatedAt"=$2 WHERE "id" = $3
===== auth code consumed ====  AuthorizationCode {
  exp: 1626257325,
  iat: 1626256725,
  nonce: '91446a375caaca3dc3573c1ef5e73934',
  scope: 'openid email access_plans tenant ',
  grantId: 'WXeSxAKj19RMwkZfh6a-78QlMgypjMv9XLOH_nsvD3T',
  authTime: 1626256608,
[...]

PR #1066, opened 2021-07-24
feat: support passing a function for cookies names configuration

JonnyBGod:
@panva Let me know if you agree with this functionality.

It is passing all existing tests but I did not add any new tests for the new behavior. Will add tests if you agree with this PR.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -5,7 +5,7 @@ will have to configure your instance with how to find your user accounts, where 
 persisted data from and where your end-user interactions happen. The [example](/example) application
 is a good starting point to get an idea of what you should provide.
 
-> ⚠️⚠️ This page now describes oidc-provider version v7.x documentation. See 
+> ⚠️⚠️ This page now describes oidc-provider version v7.x documentation. See
 [here](https://github.com/panva/node-oidc-provider/blob/v6.x/docs/README.md) for v6.x.
 
 ## Support
@@ -91,7 +91,7 @@ Since oidc-provider only comes with feature-less views and interaction handlers 
 those in, here is how oidc-provider allows you to do so:
 
 When oidc-provider cannot fulfill the authorization request for any of the possible reasons (missing
-user session, requested ACR not fulfilled, prompt requested, ...) it will resolve the 
+user session, requested ACR not fulfilled, prompt requested, ...) it will resolve the
 [`interactions.url`](#interactionsurl) helper function and redirect the User-Agent to that url. Before
 doing so it will save a short-lived "interaction session" and dump its identifier into a cookie scoped to the
 resolved interaction path.
@@ -288,7 +288,7 @@ provider.use(async (ctx, next) => {
 The following snippets show how a provider instance can be mounted to existing applications with a
 path prefix `/oidc`.
 
-Note: if you mount oidc-provider to a path it's likely you will have to also update the 
+Note: if you mount oidc-provider to a path it's likely you will have to also update the
 [`interactions.url`](#interactionsurl) configuration to reflect the new path.
 
 ### to a `connect` application
@@ -476,43 +476,43 @@ location / {
 <!-- START CONF OPTIONS -->
 ### adapter
 
-The provided example and any new instance of oidc-provider will use the basic in-memory adapter for storing issued tokens, codes, user sessions, dynamically registered clients, etc. This is fine as long as you develop, configure and generally just play around since every time you restart your process all information will be lost. As soon as you cannot live with this limitation you will be required to provide your own custom adapter constructor for oidc-provider to use. This constructor will be called for every model accessed the first time it is needed. The API oidc-provider expects is documented [here](/example/my_adapter.js).   
-  
+The provided example and any new instance of oidc-provider will use the basic in-memory adapter for storing issued tokens, codes, user sessions, dynamically registered clients, etc. This is fine as long as you develop, configure and generally just play around since every time you restart your process all information will be lost. As soon as you cannot live with this limitation you will be required to provide your own custom adapter constructor for oidc-provider to use. This constructor will be called for every model accessed the first time it is needed. The API oidc-provider expects is documented [here](/example/my_adapter.js).
+
 
 <a id="adapter-mongo-db-adapter-implementation"></a><details><summary>(Click to expand) MongoDB adapter implementation</summary><br>
 
 
-See [/example/adapters/mongodb.js](/example/adapters/mongodb.js)  
+See [/example/adapters/mongodb.js](/example/adapters/mongodb.js)
 
 
 </details>
 <a id="adapter-redis-adapter-implementation"></a><details><summary>(Click to expand) Redis adapter implementation</summary><br>
 
 
-See [/example/adapters/redis.js](/example/adapters/redis.js)  
+See [/example/adapters/redis.js](/example/adapters/redis.js)
 
 
 </details>
 <a id="adapter-redis-w-re-json-adapter-implementation"></a><details><summary>(Click to expand) Redis w/ ReJSON adapter implementation</summary><br>
 
 
-See [/example/adapters/redis_rejson.js](/example/adapters/redis_rejson.js)  
+See [/example/adapters/redis_rejson.js](/example/adapters/redis_rejson.js)
 
 
 </details>
 <a id="adapter-default-in-memory-adapter-implementation"></a><details><summary>(Click to expand) Default in-memory adapter implementation</summary><br>
 
 
-See [/lib/adapters/memory_adapter.js](/lib/adapters/memory_adapter.js)  
+See [/lib/adapters/memory_adapter.js](/lib/adapters/memory_adapter.js)
 
 
 </details>
 
 ### clients
 
-Array of objects representing client metadata. These clients are referred to as static, they don't expire, never reload, are always available. In addition to these clients the provider will use your adapter's `find` method when a non-static client_id is encountered. If you only wish to support statically configured clients and no dynamic registration then make it so that your adapter resolves client find calls with a falsy value (e.g. `return Promise.resolve()`) and don't take unnecessary DB trips.   
- Client's metadata is validated as defined by the respective specification they've been defined in.   
-  
+Array of objects representing client metadata. These clients are referred to as static, they don't expire, never reload, are always available. In addition to these clients the provider will use your adapter's `find` method when a non-static client_id is encountered. If you only wish to support statically configured clients and no dynamic registration then make it so that your adapter resolves client find calls with a falsy value (e.g. `return Promise.resolve()`) and don't take unnecessary DB trips.
+ Client's metadata is validated as defined by the respective specification they've been defined in.
+
 
 
 _**default value**_:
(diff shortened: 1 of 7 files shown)


JonnyBGod:
I am using this to enable Multi Tenancy but it could also be helpful for other use cases.

panva:
Referring back to https://github.com/panva/node-oidc-provider/discussions/916#discussioncomment-409641

> I agree this is helpful in some scenarios but i have doubts about this being the way to go. Seeing the different places you need to pull the client reference from in your code assures me of it.

> I can work from my fork for the time being not to be stuck, it would just give some comfort if we know that you are willing to consider solving this use case in the foreseeable future, so we don't invest in something the ultimately will not be supported.

I would love to have something like this to solve your needs as well as #1065, #1054, and possibly others but this is simply not it, it does not strike me as something that ticks all the boxes when it comes interplay with the rest of the features.

What I have in mind is to offer a full fledged concurrent session support with possibility to offer account chooser, etc. That needs more than just this. If there's need for it and your companies are willing to sponsor the development - drop me a note in my email inbox. Until then it has to wait until I have the time, energy and motivation to design and develop it.

JonnyBGod:
This is not to solve Multi Tenancy per say just a simple improvement to enable you to dynamically set the session names simple enough to maintain.
Even though it is not the best possible solution for multi tenancy it works and it is better than having nothing.

I have been inquired from other users about multi tenancy and how I solved it. Also seen at least a few other users inquiring about it in here as well. So it seams there is some need for it.

But again, this particular PR is quite simple a generic feature that does not add any more complexity, if you look to the changes is really quite a simple change.

And even if we need to wait for a full fledged system, having some alternative working version for now and based on such a small change to the library It seams reasonable to me.

My company I believe is sponsoring you check Fullcube.

JonnyBGod:
For reference:
https://github.com/panva/node-oidc-provider/issues/1047
https://github.com/panva/node-oidc-provider/discussions/1054
https://github.com/panva/node-oidc-provider/discussions/1037

These are all issues and discussion asking about multi tenancy aked by others.

panva:
> And even if we need to wait for a full fledged system, having some alternative working version for now and based on such a small change to the library It seams reasonable to me.

If not used for "multi tenancy" (it means to a lot of different things to a lot of people) I don't actually see use for this, hence as already said before this does not cover the feature set I would have in mind - one that starts with concurrent sessions, scoping sessions' applicability, allowing session selection prompts, ... you see where I'm going at?

> My company I believe is sponsoring you check Fullcube.

A brief sponsorship (it was cancelled after 3 weeks) is not the same as funding feature work that one needs for their deployment. It also does not imply that feature work done without any collaboration in the first place would be simply be accepted.

JonnyBGod:
> - one that starts with concurrent sessions, scoping sessions' applicability, allowing session selection prompts, ... you see where I'm going at?

Concurrent sessions is what I am trying to accomplish and I can do it with this change. The only other thing you need is a way to resolve the clientID and that's it.
Scoping sessions and prompts is also already possible with the current solutions as long as you can have concurrent sessions.
So there is not that much to do and as long as we have this change and hack a bit around resolving the clientId we can accomplish it.

You have the best lib for OICD in node space and I understand you want to focus on the core aspects of the specs, and in reality I don't think you need to maintain and support a full multi tenancy lib as this is mostly used by companies that should have the resources to maintain it.

All I am asking is to unblock the functionality with this small change.

Regarding the sponsorship, I did not know it was canceled. Will check that out. We can also discuss how we can finance your work better as well.

MichaelRo:
Adding to that, this solution is the exact solution we thought about, and probably will use
Can we add this code hidden by a feature?

panva:
Circling back to https://github.com/panva/node-oidc-provider/pull/1066#issuecomment-886099234 please respect my decision not to accept this change. The same PR has already been closed before this one and nothing has changed since.

There will be native support for concurrent sessions, scoping sessions' applicability and account choosing eventually. The "when" depends on either my free time and enthusiasm about the feature (which is low atm) or someone coming in and funding its development.

Over the 6 years and 7 major versions the library has grown considerably. Single-handedly maintaining it, improving it, continuously certifying it, etc is not a trivial task. Adding features needs a bigger picture first approach. Accepting even a small change needs to have interplay with the rest of the features (existing and future ones).

JonnyBGod:
Well with this kind of response it is not easy to break from the "Single-handedly maintaining it" situation, it it?

There are at least 3 developer here that need this and are willing to help. Just repeating the money problem and the lack of time does not solve anything does it?

We proposed a plan that works for the interested parties. Can you share your plan with us, we might be able to help...

You ask us to respect your decision but I don't see a lot of respect for our needs and contributions when the repeated response is "No because I don't want it and have a plan...".

Help us understand your plan and I will help develop it.
Its your library and you decide, but don't ask me to respect your decision when you give me no respectable answers.

Also the previous PR for version 6 was too hacky and at the time I understood why you did not want to include it. But now with version 7 the needed changes are very minor and do not create any overhead. That is why I proposed a PR again.
This time I do not understand why you don't consider it.

Again, tell me how you see this being done and I will do it.

panva:
> Well with this kind of response it is not easy to break from the "Single-handedly maintaining it" situation, it it?

It was not a complaint, but a statement. I'm not attempting to break from it.

Now, just now, did you ask the first time "how do I want it?", before then it was only "can we please have this thing?".

> Help us understand your plan

I will unblock your needs to drive the cookie name value dynamically through an OIDCContext prototype method. It will maybe be only done for the `'session'` cookie, but you're then free to overload it and do whatever whacky and hacky resolution that you have in this proposed helper today. It will not be part of the public documented API and whenever a full blown solution is there it will most likely be completely decommissioned. 

I welcome those who need the features described (concurrent sessions, scoping sessions' applicability, allowing session selection prompts) to sponsor the library and get a chance to be a part of the design process for a native solution then.

JonnyBGod:
Ok fair enough.

We just need the session cookies for this.

xelinorg:
Usually I do not like to solve problems for others when I have not solved my own. When I settle and I have decided what is the plan for solving the problem,  I take ownership and go ahead to support my decision and if there is delegation or if this could effect others  I always think twice and I am not an easy go... 

@JonnyBGod you could fork and take the responsibility of your code and what you have done to solve the multi-tenancy problem or any other problem your solution addresses. If you offload here then all of us who are using the library could have issues and is not very sure that you will be here to support... @panva has created this and is doing the support for some years now so I expect him to be around.

panva:
@JonnyBGod see the `dynamic-session-cookie-name` branch and tell me if it works for you.

```js
const provider = new Provider(...);
const fn = function () {
  const { ctx } = this;
  return 'foobar';
};
Object.defineProperty(provider.OIDCContext.prototype, 'sessionCookieName', {
  get: fn,
});
```

JonnyBGod:
@panva something like this can work but would need to support returning a promise or having a call back as we need to make a request to the database.

panva:
What do you need to fetch from the database that isn't otherwise available?

JonnyBGod:
We are correlating oidcClients with "organizations" and currenlty setting the session cookie names with the "organization" id.

Now that I think about it I think we can just use the client id instead and resolve Organization later.

There is still a use case that I am not sure it is still needed since version 7 though.

This is the way we try to resolve the clientId:
```
async getClientId(ctx): Promise<string> {
    let clientId = ctx.oidc?.client?.clientId || ctx.oidc.client?.id

    if (!clientId && ctx.request.query['client_id']) {
      clientId = ctx.request.query['client_id'] as string
    }

    if (!clientId && ctx.request.headers?.referer) {
      const query = new URL(ctx.request.headers.referer).searchParams
      clientId = query['client_id'] as string
    }

    if (!clientId && ctx.request.query['id_token_hint']) {
      clientId = decode(ctx.request.query['id_token_hint'] as string)?.payload?.aud
    }

    if (!clientId && ctx.oidc) {
      const id = ctx.oidc.cookies.get('_resume') || ctx.oidc.cookies.get('_interaction')

      if (id) {
        const {
          payload: { params },
        }: // eslint-disable-next-line @typescript-eslint/no-explicit-any
        any = await this.collection.findOne({ _id: id })

        if (!params?.client_id) {
          throw new Error('interaction session id cookie not found')
        }

        clientId = params.client_id
      }
    }

    return clientId
}
```

At the time we needed all these steps to support all the possible flows. I did not yet checked the need for these since upgrading to version 7 though. If we can live without the last snerario there, we would not need async support anymore.

panva:
@JonnyBGod So, can you work with just client_id synchronously? Or keep the overall mapping between client_id and organization in process cached and prepared to be used synchronously? Is that an unsolvable problem if this remains sync as it is today?

JonnyBGod:
While not ideal I could live with a solutions such as a in memory cache for clientId -> organizationId.

However I have yet to find a solution when we need to resolve the clientId from a resume or interaction cookie. In this case we still need to grab the cookie and get it from the database.

If you see some options here I would be happy to try it. Otherwise I think we need the async functionality.

JonnyBGod:
Also happy to ear if you have any other possible solutions for the session naming.
We think the main requisites are to enable parallel sessions between clients while disallowing multiple session for the same client or in our case for the same organization.

panva:
I see what you mean, I would suggest you to store the organization or client in your own cookie to read during interaction resume. You can do so e.g. in the `interactions.url` helper through [`ctx.cookies.set(name, value, [options])
`](https://koajs.com/#ctx-cookies-set-name-value-options-).

JonnyBGod:
Ok seams like a solution.
Will have to try it out for all use cases.
Give me a couple of days to try this approach.

JonnyBGod:
I believe the same problem with resolving the Client Id persists.

The new cookie would need to correlate Client Id with Organization ID we would still need to resolve the Client Id on "sessionCookieName" call.

It's not difficult to solve the Organization Id problem, the whole issue from the beginning is related to resolving Client ID for all possible interaction flows.

panva:
The mapping is something you gotta figure out and keep available.

When you're on `ctx.oidc.route === 'authorize'` the client id is resolvable synchronously. When interactions are required you set a cookie for yourself in `interactions.url`.

When you're on `'resume'` you read that cookie and use that.

I don't see an issue. You may as well have your client identifiers contain the organization identifier.

panva:
You can also pass the client id or organization identifier along to your resume route as a query argument.

JonnyBGod:
I got to the point where I create a new cookie as you proposed with name set to "org_INTERACTION_UID", then on "sessionCookieName" I can resolved the cookie from the interaction uid that I grab from the context params.

This works fine during most interactions and I managed to complete the login flow.

The problem I got then was when login in again with an existing session.
In this case the context does not have an interaction uid meaning that I would need to fall back to a client id based session cookie name.

And for that I need to use this async function:
```typescript

async function getClientId(ctx, collection: Collection<AdapterPayload>): Promise<string> {
  let clientId = ctx.oidc?.client?.clientId || ctx.oidc.client?.id

  if (!clientId && ctx.request.query['client_id']) {
    clientId = ctx.request.query['client_id'] as string
  }

  if (!clientId && ctx.request.headers?.referer) {
    const query = new URL(ctx.request.headers.referer).searchParams
    clientId = query['client_id'] as string
  }

  if (!clientId && ctx.request.query['id_token_hint']) {
    clientId = decode(ctx.request.query['id_token_hint'] as string)?.payload?.aud
  }

  if (!clientId && ctx.oidc) {
    const id = ctx.oidc.cookies.get('_resume') || ctx.oidc.cookies.get('_interaction')

    if (id) {
      const {
        payload: { params },
      }: // eslint-disable-next-line @typescript-eslint/no-explicit-any
      any = await collection.findOne({ _id: id })

      if (!params?.client_id) {
        throw new Error('interaction session id cookie not found')
      }

      clientId = params.client_id
    }
  }

  return clientId
}
```

Any other ideas?

panva:
I don't understand. Assuming client id can be resolved to a bucket.

- at /authorize I can always get the bucket because you always have a client_id
- when interactions are required interactions.url is called and you store the bucket
- at the "resume" endpoint I can always get the bucket because it is in the cookie

Nevertheless

- you can make your client identifiers structured to always have the bucket in its value
- you can pass the bucket to resume without a cookie using a query string parameter

JonnyBGod:
I believe it is working.

```typescript
interactions: {
  ...configuration.interactions,
  url: async (ctx: KoaContextWithOIDC, interaction): Promise<string> => {
    const clientId = getClientId(ctx)

    if (clientId) {
      const iv = randomBytes(16)
      const cipher = createCipheriv(algorithm, secretKey, iv)

      ctx.cookies.set(
        // eslint-disable-next-line @typescript-eslint/no-explicit-any
        `_client_${interaction.uid || (ctx.oidc as any)?.uid}`,
        `${Buffer.concat([cipher.update(clientId.toString()), cipher.final()]).toString('hex')}:${iv.toString(
          'hex'
        )}`,
        {
          maxAge: 3600 * 1000, // 1 hour in milliseconds
          overwrite: true,
          signed: true,
        }
      )
    }

    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    return `/interaction/${interaction.uid || (ctx.oidc as any)?.uid}`
  },
},
```

```typescript
this.provider.on('interaction.ended', (ctx: KoaContextWithOIDC) => {
  if (ctx.params.uid) {
    ctx.cookies.set(`_client_${ctx.params.uid}`, null)
    ctx.cookies.set(`_client_${ctx.params.uid}.sig`, null)
  }
})

Object.defineProperty(this.provider.OIDCContext.prototype, 'sessionCookieName', {
  get: function () {
    const { ctx } = this
    let clientId = ctx.oidc?.client?._id

    if (!clientId) {
      clientId = getClientId(ctx)
    }

    if (clientId) {
      return `_session_${createHash('md5').update(clientId.toString()).digest('hex')}`
    }

    return null
  },
})
```

```typescript
function getClientId(ctx): string {
  let clientId = ctx.oidc?.client?.clientId || ctx.oidc.client?.id

  if (!clientId && ctx.request.query['client_id']) {
    clientId = ctx.request.query['client_id'] as string
  }

  if (!clientId && ctx.request.headers?.referer) {
    const query = new URL(ctx.request.headers.referer).searchParams
    clientId = query['client_id'] as string
  }

  if (!clientId && ctx.request.query['id_token_hint']) {
    clientId = decode(ctx.request.query['id_token_hint'] as string)?.payload?.aud
  }

  if (!clientId && ctx.params.uid) {
    const cookies = getCookiesAsCollection(ctx.request.header.cookie)
    // eslint-disable-next-line no-prototype-builtins
    if (cookies.hasOwnProperty(`_client_${ctx.params.uid}`)) {
      const [hash, iv] = cookies[`_client_${ctx.params.uid}`].split(':')
      const decipher = createDecipheriv(algorithm, secretKey, Buffer.from(iv, 'hex'))
      clientId = Buffer.concat([decipher.update(Buffer.from(hash, 'hex')), decipher.final()]).toString()
      ctx.cookies.set(`_client_${ctx.params.uid}`, null)
      ctx.cookies.set(`_client_${ctx.params.uid}.sig`, null)
    }
  }

  return clientId
}
```

Will keep testing but untill now every flow worked including "recover account/reset password", "third party passport login", "email verification".

Quick question, do you see any better method to clear the cookies when interaction cookies are cleared?

JonnyBGod:
@panva Everything is working fine for a week now with this new feature.
Will you publish it or keep it on a separate branch?

PR #1093, opened 2021-09-10
example: certification code typo

btsheehy:
Just stumbled across this typo where `ctx.stats` should be `ctx.status`


coltonwood (approved):
Nice catch 👍

PR #1124, opened 2021-10-01
Increase MAX_TTL of PAR object

siqueirarenan:
Hey,
I would like to increase the TTL of the PAR response and I found that this is hard-coded. As per specification (https://www.rfc-editor.org/rfc/rfc9126.html) this value could be up to 600 (or more, if desired). Can we make this adjustment while this feature is not added to the configuration file?
Thanks


panva:
Why?

siqueirarenan:
hahha, yah it is indeed. In our case they want to implement a handoff redirect, where the redirect_uri would be available in one device though e. g. QR code, and the person needs some time to take the other device and scan it. UX folks claim that 60s might not be enough. But yah, the client can always push a new request right...

PR #1150, opened 2021-11-28
fix: set paseto token config on ResourceServer

dev-Fabi:
Currently using `paseto` as access token type does not work, as the paseto config from the resourceServer object returned by `getResourceServerInfo` is not copied/set in the ResourceServer constructor.


panva:
Landed in 02c821d, thank you @dev-Fabi

PR #1177, opened 2022-01-07
fix: slice > substr(ing) in mountPath

qfdk:
- rollback slice to substring in oidc_context.js


panva:
thank you for spotting this regression and opening a PR to fixing it, if you could come up with a unit test in a followup PR that'd be great.

qfdk:
@panva If i have some time to do that i will have a try.

PR #1185, opened 2022-02-21
example: update skip_consent recipe

chrisskilton:
…sion expiry

* currently if a session gets re-established without an explicit logout, the grantId gets copied onto the new session but its expiry does not get updated
* this means the grant can expire before the session and the interactionPolicy will try and prompt for consent
* this code snippet inspects the expiry on the session and updates the grant to match if the values differ


panva (changes requested):

panva:
Great that you figured it out and thank you for submitting this update.

PR #1187, opened 2022-03-12
Fix function name

francescopersico:
I was inspecting the code and i noticed the name of the function was not aligned.
The fix align the function name.


panva:
Thank you.

PR #1188, opened 2022-03-17
feat: configurable ignoring of offline access scope

charsleysa:


charsleysa:
Closed due to different solution for issue.

PR #1192, opened 2022-05-16
fix: updated signed to trusted in interaction model

davidgtonge:
The key for signed params was changed from `signed` to `trusted`, however unfortunately the interaction model wasn't updated and so the trusted/signed params list isn't being persisted in the interaction session.


panva:
Thank you Dave!

PR #1193, opened 2022-05-30
Allow issuer without trailing /

mitar:
When issuer does not have trailing `/`, `pathname` is null.


mitar:
Interesting. That was happening only because I was using custom scheme.

PR #1196, opened 2022-06-14
fix(configuration): enabledJWA validation

javier-garcia-meteologica:
Validator should be excluding HS and none algorithms from enabledJWA instead of enforcing them

### DIFF
--- a/lib/helpers/configuration.js
+++ b/lib/helpers/configuration.js
@@ -6,6 +6,7 @@ const remove = require('./_/remove');
 const merge = require('./_/merge');
 const pick = require('./_/pick');
 const set = require('./_/set');
+const without = require('./_/without');
 const formatters = require('./formatters');
 const docs = require('./docs');
 const getDefaults = require('./defaults');
@@ -276,7 +277,7 @@ class Configuration {
   defaultSigAlg() {
     const allowList = this.enabledJWA;
 
-    this.setAlgs('idTokenSigningAlgValues', allowList.idTokenSigningAlgValues.filter(filterHSandNone));
+    this.setAlgs('idTokenSigningAlgValues', without(allowList.idTokenSigningAlgValues, filterHSandNone));
     this.setAlgs('idTokenEncryptionAlgValues', allowList.idTokenEncryptionAlgValues.slice());
     this.setAlgs('idTokenEncryptionEncValues', allowList.idTokenEncryptionEncValues.slice(), 'encryption.enabled');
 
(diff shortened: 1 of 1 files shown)


panva (changes requested):
I don't understand why this change would be needed, would you please explain and, most importantly, add test cases?

panva:
https://github.com/panva/node-oidc-provider/actions/runs/2497928004

This PR removes the ability to enable the HMAC and none algorithms altogether. I'm closing this since this functionality works as intended. Through the `enabledJWA` and `jwks` configuration one can enable any of the supported algorithms.

javier-garcia-meteologica:
For these `enabledJWA` settings, the default algorithms were asymmetric but the all of them are excluded because the filter only accepts symmetric algorithms or none, which seems inconsistent.

https://github.com/panva/node-oidc-provider/blob/f5c9a3f9d4bd83df24959e7bfc5d354a5015164a/lib/helpers/defaults.js#L2462-L2464
https://github.com/panva/node-oidc-provider/blob/f5c9a3f9d4bd83df24959e7bfc5d354a5015164a/lib/helpers/defaults.js#L2504-L2506
https://github.com/panva/node-oidc-provider/blob/f5c9a3f9d4bd83df24959e7bfc5d354a5015164a/lib/helpers/defaults.js#L2525-L2527
https://github.com/panva/node-oidc-provider/blob/f5c9a3f9d4bd83df24959e7bfc5d354a5015164a/lib/helpers/defaults.js#L2545-L2547

Let's take a look to `enabledJWA.idTokenSigningAlgValues`. From [OpenID Connect Discovery 1.0 Spec](https://openid.net/specs/openid-connect-discovery-1_0.html)

> id_token_signing_alg_values_supported
    REQUIRED. JSON array containing a list of the JWS signing algorithms (alg values) supported by the OP for the ID Token to encode the Claims in a JWT [[JWT]](https://openid.net/specs/openid-connect-discovery-1_0.html#JWT). The algorithm RS256 MUST be included. The value none MAY be supported, but MUST NOT be used unless the Response Type used returns no ID Token from the Authorization Endpoint (such as when using the Authorization Code Flow).

Given that there is a conflict between the default values `['RS256', 'PS256', 'ES256', 'EdDSA']` and the filtered ones `['HS*', 'none']`, the spec seems to favor the former and discourage the later.

In general symmetric algorithms are discouraged in favor of asymmetric algos
https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/#RSA-or-HMAC-

That's why I think the filter rule was a bug and it intended to do the opposite.

Also, when I started using this library it would always throw this error

```
error: invalid_client_metadata
error_description: id_token_signed_response_alg must not be provided (no values are allowed)
```

Which is caused because the filter inconsistency makes `enabledJWA.idTokenSigningAlgValues` to always be empty.

javier-garcia-meteologica:
Now I've realized my mistake, you were right. The `idTokenSigningAlgValues` that was giving me problems was part of `initialize_keystore`. `node-oidc-provider` is correctly validating allowed algos, the real validation of asymmetric algos happens when the jwks are loaded.

The problem was caused because I didn't generate a new keyset as stated in the example. I skipped that step and used a preexisting keyset containing a single RSA key with `"alg": "PS512"`. `node-oidc-provider` reads the `alg` header and restricts its usage to that algorithm which by default is not allowed. In this case I needed to:

- Declare the extra algo in `client.id_token_signed_response_alg`
- Extend `enabledJWA.idTokenSigningAlgValues` with extra algos
- If compatibility with the rest of RSA signing algorithms (`RS256`, `PS256`, `RS512`) is desired, the property `alg` has to be dropped from the RSA key

PR #1197, opened 2022-07-01
Feature: allow arrow-function & class-methods as adapter factory

talentumtuum:
Hi!
Now when implementing the "AdapterFactory" type from the package https://www.npmjs.com/package/@types/oidc-provider
a situation can happen where the types are correct, but at runtime we get an error.

I found a mention of this problem here https://github.com/panva/node-oidc-provider/pull/776#issuecomment-714742339

With this PR I add the ability to use arrow-functions and class methods as a factory and at the same time fix the non-obvious behavior when writing code in typescript

### DIFF
--- a/lib/helpers/initialize_adapter.js
+++ b/lib/helpers/initialize_adapter.js
@@ -1,7 +1,10 @@
+const util = require('util');
+
 const MemoryAdapter = require('../adapters/memory_adapter');
 
 const instance = require('./weak_cache');
 const attention = require('./attention');
+const { isConstructable } = require('./type_validators');
 
 module.exports = function initializeAdapter(adapter = MemoryAdapter) {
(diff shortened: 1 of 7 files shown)


co1dmoon:
like

talentumtuum:
@panva Ready for review

talentumtuum:
@panva I'll take care of this problem in my spare time, I think I'll be back with a solution in 1-3 days, maybe today. Okay?

panva:
Thank you, for sure, come back to the problem any time. FWIW it might be as simple as updating `base_model.js` where the adapters get currently instantiated.

talentumtuum:
@panva Thanks, Thank you for your quick assistance

panva:
Thank you for the PR!

PR #1200, opened 2022-08-19
Add sub claim in client credentials grant

vaibhav-medhekar:
Hi All,

We have a requirement to add 'sub' claim in the jwt token for client credentials grant type.
The value of the 'sub' claim will be the clientId.

The specification mentions below details for client credentials grant:
"For client authentication, the subject MUST be the "client_id" of the OAuth client."

The specification link for the same is as below:
https://datatracker.ietf.org/doc/html/rfc7523#section-3

Could you please review and merge this MR. Please let us know if any review comments.

Thanks & Regards,
Vaibhav Medhekar

### DIFF
--- a/lib/models/client_credentials.js
+++ b/lib/models/client_credentials.js
@@ -14,6 +14,7 @@ module.exports = (provider) => class ClientCredentials extends apply([
       'aud',
       'extra',
       'scope',
+      'accountId'
     ];
   }
 };


panva:
This PR does not a whole lot of sense to me. The specification is effectively for private_key_jwt client authentication, where the AS does not issue the assertion, the client does, and there `sub` is required to be the client_id.

And issued JWT Access Tokens for the client credentials grant already have the `sub` value populated by the client_id.

vaibhav-medhekar:
@panva 
Following is the decoded JWT for the client credentials grant and it does not contain the `sub` claim in it. Instead the client_id is stored in `aud` claim. Because of this issue, we wanted enhancement so that we could have `sub` claim in client credentials grant.

image

panva:
You are not using the v7.x major then. Make sure to use the latest, I cannot accept your PR.

PR #1201, opened 2022-09-14
test,docs: updated fastify versions

marco-ippolito:
Hi,
I wanted to add a working example on fastify since it seems there isn't one, and I've also bumped the fastify version to a more recent one.


panva (changes requested):
This example does not use the routes or configurations from the examples folder. I believe you could just update the [Mounting oidc-provider](https://github.com/panva/node-oidc-provider/blob/main/docs/README.md#to-a-fastify-application) with up to date information which is then part of the CI run in `test/test_helper.js`

marco-ippolito:
> Thank you, the fastify content could definitely use a refresher.
> 
> This example does not use the routes or configurations from the examples folder. I believe you could just update the [Mounting oidc-provider](https://github.com/panva/node-oidc-provider/blob/main/docs/README.md#to-a-fastify-application) with up to date information which is then part of the CI run in `test/test_helper.js`

Ok I've removed the example maybe it would be nice to have one working like the others.

docs/README.md
```diff
@@ -300,12 +300,20 @@ connectApp.use('/oidc', oidc.callback());
 
 ### to a `fastify` application
 ```js
-// assumes fastify ^3.0.0
-await app.register(require('fastify-express'));
-// or
-// await app.register(require('middie'));
-
-fastifyApp.use('/oidc', oidc.callback());
+// assumes fastify ^4.5.0
```
simoneb, line 303: nit: fastify^4 is good enough

docs/README.md
```diff
@@ -300,12 +300,20 @@ connectApp.use('/oidc', oidc.callback());
...
+async function build() {
+  const fastify = Fastify();
+  await fastify.register(require('@fastify/express'));
+  // or
+  // await app.register(require('middie'));
```
simoneb, line 308: @fastify/middie

panva:
@marco-ippolito @simoneb thank you for this PR. I took the liberty of fixing it up a bit and will merge it like so.


> it would be nice to have one working like the others.

I would be happy to have a fastify example should it have the same functionality as the koa, express, and standalone ones.

PR #1202, opened 2022-09-19
fix: skip_consent recipe causing "TypeError: Cannot read property 'ex…

rdubigny:
…p' of undefined"

At the end of the request processing, session.exp is set to 'currentTime + sessionTtl' in sessionHandler. This result in grant.exp to be set to previous session.exp, not future session.exp.

This end in various bugs :

1. `TypeError: Cannot read property 'exp' of undefined` (line 28) as grantId is still referenced in Session but unreachable because it has expired
2. grant.exp might be equal to ctx.oidc.session.exp but as ctx.oidc.session.exp will be updated later on the test `grant.exp < ctx.oidc.session.exp` is irrelevant.

To fix this, I propose to remove the irrelevant test and to set grant.exp to 'currentTime + sessionTtl'

### DIFF
--- a/recipes/skip_consent.md
+++ b/recipes/skip_consent.md
@@ -25,8 +25,11 @@ const oidcConfiguration = {
       // this aligns the Grant ttl with that of the current session
       // if the same Grant is used for multiple sessions, or is set
       // to never expire, you probably do not want this in your code
-      if (ctx.oidc.account && grant.exp < ctx.oidc.session.exp) {
-        grant.exp = ctx.oidc.session.exp;
+      if (ctx.oidc.account) {
+        // Change sessionTtl to match the value set for oidc's "Session.ttl" in your configuration
+        const sessionTtl = 14 * 24 * 60 * 60; // 14 days in seconds, the default session ttl
+        const epochCurrentTime = Math.floor(Date.now() / 1000);
+        grant.exp = epochCurrentTime + sessionTtl;
 
         await grant.save();
       }
--- a/recipes/skip_consent.md
+++ b/recipes/skip_consent.md
@@ -13,28 +13,25 @@ To achieve that you want to add the requested claims/scopes/resource scopes to t
 ```js
 const oidcConfiguration = {
   loadExistingGrant(ctx) {
+    // keep grant expiry aligned with session expiry
+    // to prevent consent prompt being requested when grant expires
+    await Promise.all(
+      Object.entries(ctx.oidc.session.authorizations).map(async ([, { grantId }]) => {
+        const grant = await ctx.oidc.provider.Grant.find(grantId);
+        // this aligns the Grant ttl with that of the current session
+        // if the same Grant is used for multiple sessions, or is set
+        // to never expire, you probably do not want this in your code
+        grant.exp = epochTime() + sessionTtlInSeconds;
+        await grant.save();
+      })
+    );
+    
     const grantId = (ctx.oidc.result
       && ctx.oidc.result.consent
       && ctx.oidc.result.consent.grantId) || ctx.oidc.session.grantIdFor(ctx.oidc.client.clientId);
 
     if (grantId) {
-      // keep grant expiry aligned with session expiry
-      // to prevent consent prompt being requested when grant expires
-      const grant = await ctx.oidc.provider.Grant.find(grantId);
-
-      // this aligns the Grant ttl with that of the current session
-      // if the same Grant is used for multiple sessions, or is set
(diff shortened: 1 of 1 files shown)


panva:
Hi @rdubigny, can you expand on what this is fixing exactly? How can I reproduce, what are the full stack traces, etc.

rdubigny:
Hi @panva, thanks for your quick feedback.

I use this recipe to skip consent screen. But I got error `TypeError: Cannot read property 'exp' of undefined` on my production environement.

After digging in the logs I found out that when users start a login flow form RP while they still got a living session on my oidc server, the Session entry in redis was extended but the Grant entry was not. The rest of the investigation is described in my previous comment.

Step to reproduce : 

1. set ttl of Grant and Session to 10 minutes
2. login from RP
3. wait for 6 minutes (at this point, Session age is 4 minutes, Grant age is 4 minutes)
4. reset session on RP (ex: delete cookies) but keep it on oidc server
5. login from RP (at this point, Session age is 10 minutes, Grant age is 4 minutes)
6. wait for 6 minutes (at this point, Session age is 4 minutes, Grant has expired)
7. reset session on RP (ex: delete cookies) but keep it on oidc server
8. login from RP => this trigger the error since Session refers to an expired Grant (at this point, Session age is 10 minutes, Grant has expired)

Here is the stack trace in debug mode : 

```
  oidc-provider:server_error:trace TypeError: Cannot read property 'exp' of undefined
  oidc-provider:server_error:trace     at _callee3$ (current/src/oidc-provider-configuration.js:57:37)
  oidc-provider:server_error:trace     at tryCatch (current/node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:65:40)
  oidc-provider:server_error:trace     at Generator.invoke [as _invoke] (current/node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:303:22)
  oidc-provider:server_error:trace     at Generator.prototype.<computed> [as next] (current/node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:117:21)
  oidc-provider:server_error:trace     at step (current/src/oidc-provider-configuration.js:18:191)
  oidc-provider:server_error:trace     at current/src/oidc-provider-configuration.js:18:361
  oidc-provider:server_error:trace     at processTicksAndRejections (internal/process/task_queues.js:97:5) +0ms
```

`current/src/oidc-provider-configuration.js:57:37` refers to the line `if (ctx.oidc.account && grant.exp < ctx.oidc.session.exp) {` of the recipe.

Make me know if you need more information. I will be glad to help.

rdubigny:
I have implemented a more resilient fix on my own project : https://github.com/betagouv/api-auth/commit/632f7ab24a6c21922c92b28cd01c6367d128246b

The problem also occured when the user used multiple RPs simultaneously. The new implementation now update all living Grants.

However, this implementation seems a bit dirty and unstable to me. I wish I could disable the Session extention rather than trying to extend all these tokens.

Let me know if you have any insight on this issue.

panva:
Honestly I think it's way simpler to just make the grants just not expire at all. I'm not sure if that's atm done by setting an insane ttl on the Grant model, setting it to Infinity, or 0. Not sure.

rdubigny:
Hi @panva,

I monitored my new implementation in production for a week. No error was thrown so I'll be keeping this solution for now.

My opinion is that this PR propose a more stable implementation than the one suggested in the original recipe. I does not understand the big picture though. I let you decide if you rather merge this improvment or suggest to set a very long ttl for Grant tokens.

rdubigny:
I recently went into new errors on this (which are very rare now but still). I have the felling that setting an infinit ttl for Grant model will postpone the problem instead of fixing it. I was wondering if we could just create a new Grant at each interaction instead of reusing previous grants.

@panva do you thinks this could work without twisting this lib usage too much ?

waynerobinson:
I agree with @panva, most systems (outside of open banking) issue permanent grants until they're revoked by the user.

We'd use `Infinity`, but `checkTTL()`'s use of `Number.isSafeInteger(value)` returns `false` for `Infinity` for good reason as most DB layers won't have a way to handle `Infinity` in their expiry attribute.

Currently we set the Grant's expiry to 100 years in the future (we're using DynamoDB which supports numbers up to 38 significant digits), which works fine for us currently.

panva:
I'm going to go ahead and close this because

a) i don't believe one session's grant ttl refresh should affect that of other clients in the same session.
b) if you need non-expiring grants, make them non-expiring (or at the very least not expiring in the near future).

PR #1203, opened 2022-09-21
Remove Assertion of Request/Response for getInteraction()

waynerobinson:
Running assertions every request reduces performance and prevents requests/responses that match the shape required by the Provider (but not matching the specific classes) fail.

I encountered this issue whilst attempting to get oidc-provider to work with Fastify & https://github.com/fastify/aws-lambda-fastify on Lambda.

aws-lambda-fastify creates a request-like object it passes into Fastify, but this doesn't match the `instanceof` assertions in `getInteraction()`.

Discovering if the `oidc-provider` integration works with a particular set of requests or not should only be done once, during development and not every request.

### DIFF
--- a/lib/provider.js
+++ b/lib/provider.js
@@ -40,21 +40,7 @@ const models = require('./models');
 const ssHandler = require('./helpers/samesite_handler');
 const get = require('./helpers/_/get');
 
-function assertReqRes(req, res) {
-  assert(
-    req instanceof IncomingMessage || req instanceof Http2ServerRequest,
-    'first argument must be the request (http.IncomingMessage || http2.Http2ServerRequest), for express req, for koa ctx.req',
-  );
-  if (arguments.length === 2) {
-    assert(
-      res instanceof ServerResponse || res instanceof Http2ServerResponse,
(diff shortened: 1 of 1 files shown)


panva (commented):
@waynerobinson thank you

This check is in place largely because of previous issues encountered by other developers who were not passing compliant objects. This caused the operations to fail with very obscure errors that were a pain to debug. I am therefore hesitant to remove it.

I wouldn't mind a PR where the existing API doesn't change but there's a private-ish API that omits this change but otherwise the implementation is shared between the current and the private-ish one.

waynerobinson:
No problems. I'll implement something that makes this optional vs removing it altogether.

Would it make sense to change the test it self to make sure that the _shape_ of the request/response is correct rather than checking for specific instances from Koa/Express?

panva:
What it is checking for is not framework specific, it is a Node.js builtin constructor and I'm not willing to give it up.

waynerobinson:
We've changed out architecture a bit to make this irrelevant so we probably won't be implementing an optional workaround here.

Thank you for your feedback though. 👍

PR #1204, opened 2022-09-29
Replace Assert with Return

waynerobinson:
Node's native `assert()` can be very heavy, especially in a large project with a source map as it creates a full stack trace on failure.

In practice, we were seeing negative responses to `model.find()` taking up to 2 seconds to resolve.

Most assertions in this library are either for config or use Koa's `assert` method, which uses a considerably lighter code-path.

Replacing with if `(!stored) { return }` is an effectively free operation and solved our performance issues on negative responses.


panva (commented):
Move it out of the try block please.

waynerobinson:
Moved. 😀

panva:
I'll go through the rest of the hot path asserts and replace them. Since i'm not able to push to your upstream I'll do the rest of the work in #1205 where i've kept you as co-author.

waynerobinson:
No problem, thanks for that. It's a big library and I'm a little unsure what should and shouldn't be altered. 😅

PR #1205, opened 2022-10-05
perf: replace native asserts

panva:
Node's native `assert()` can be very heavy, especially in a large project with a source map as it creates a full stack trace on failure.

Closes #1204


panva:
cc @waynerobinson

PR #1208, opened 2022-11-22
fix: add code_verifier param to authorization_code grant

andrewrothman:
The function referenced the param, but it was not available unless explicitly allowed.

I think this is right but I could be missing something. First time contribution.

Seems to apply to both v7 and v6.

Thanks!
Andrew

### DIFF
--- a/lib/actions/grants/authorization_code.js
+++ b/lib/actions/grants/authorization_code.js
@@ -218,4 +218,4 @@ module.exports.handler = async function authorizationCodeHandler(ctx, next) {
   await next();
 };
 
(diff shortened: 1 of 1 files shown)


panva (changes requested):
Hi @andrewrothman 

thank you for your contribution. 

The code verifier parameter is actually added in [here](https://github.com/panva/node-oidc-provider/blob/main/lib/helpers/initialize_app.js#L108). 

Now I don't recall the historical reasons why that is so let's see what happens if you can remove it from there.

PR #1211, opened 2022-12-19
Custom confirmation parameter - Device Flow

moradfaris2:
**The issue:**
In the feature of device flow - when we use 2 parameters "confirm" and "abort" then no matter what is sent in the value of "confirm" it will confirm.
Example-> if when using "confirm" the value is false/no, it will still confirm.

The reason this is an issue for me is that I'm trying to use the library as an API solution without the "hosted" pages.

**The suggestion:**
In order to be backwards compatible I suggest adding a configuration for "useBooleanConfirmation" that will let the developer decide to use only one approval field which accepts only boolean values.

**Implementation details:**
I added configurations called `useBooleanConfirmation, confirmParamName` that lets the developer decide to use one boolean param name and the param name itself. (perhaps you will decide the second one is not needed).

### DIFF
--- a/lib/actions/code_verification.js
+++ b/lib/actions/code_verification.js
@@ -127,8 +127,13 @@ export const post = [
       const action = ctx.oidc.urlFor('code_verification');
       await userCodeConfirmSource(
         ctx,
-        formHtml.confirm(action, ctx.oidc.session.state.secret, userCode, useBooleanConfirmation,
-          confirmParamName),
+        formHtml.confirm(
+          action,
+          ctx.oidc.session.state.secret,
+          userCode,
+          useBooleanConfirmation,
+          confirmParamName,
+        ),
         client,
         code.deviceInfo,
         denormalize(normalized, mask),
--- a/lib/actions/code_verification.js
+++ b/lib/actions/code_verification.js
@@ -48,7 +48,20 @@ export const get = [
 export const post = [
   sessionMiddleware,
   parseBody,
-  paramsMiddleware.bind(undefined, new Set(['xsrf', 'user_code', 'confirm', 'abort'])),
+
+  async function codeVerificationParams(ctx, next) {
+    const { useBooleanConfirmation, confirmParamName } = instance(ctx.oidc.provider).configuration('features.deviceFlow');
+
+    const allowParamList = new Set(['xsrf', 'user_code']);
+
+    allowParamList.add(confirmParamName);
+    if (!useBooleanConfirmation) {
+      allowParamList.add('abort');
+    }
+
+    return paramsMiddleware(allowParamList, ctx, next);
+  },
+
   rejectDupes.bind(undefined, {}),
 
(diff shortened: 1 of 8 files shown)


panva:
Hello @moradfaris2 

> no matter what is sent in the value of "confirm" it will confirm.

That is incorrect. A falsy string value (empty string) will not confirm.

I believe any frontend hosted or not should be able to emulate the current form setup and thus this is not needed.

moradfaris2:
> 

Hey @panva 
Thank you for your reply.



> Hello @moradfaris2
> 
> > no matter what is sent in the value of "confirm" it will confirm.
> 
> That is incorrect. A falsy string value (empty string) will not confirm.
> 
> I believe any frontend hosted or not should be able to emulate the current form setup and thus this is not needed.

I am using the oidc-provider library as an API solution, where the frontend is not hosted by the same server as the library. Currently, when using the device flow feature with the 'confirm' and 'abort' parameters, any value passed to 'confirm' will confirm the request. This is a problem for me because I want to be able to use a boolean value (e.g. true or false) to indicate whether the request should be confirmed or not.

For example, let's say my frontend sends a request to the library with 'confirm=false'. With the current implementation, this request would still be confirmed, which is not the behavior I want. Instead, I want the request to be confirmed only if 'confirm=true' is sent.

To solve this problem, I suggest adding a configuration called 'useBooleanConfirmation' that allows the developer to specify whether the 'confirm' parameter should accept only boolean values. This would allow me to use the library as an API solution in a more flexible way, and would be backwards-compatible with the existing behavior.

I have implemented this change and added a 'confirmParamName' configuration as well, which allows the developer to specify the name of the confirmation parameter. This can be useful if the developer wants to use a different parameter name for confirmation (e.g. 'approve' instead of 'confirm').

I understand that the maintainers of the library have the final say on what changes are accepted, and I appreciate any feedback you can provide on this suggestion. Thank you!

panva:
My suggestion is that if you want to step out of the intended way this library is used you do it in a way that does not add maintenance burden on said library.

I don't think anything blocks you from achieving your way you just have to stick to the form parameters and values the way it is.

moradfaris2:
> My suggestion is that if you want to step out of the intended way this library is used you do it in a way that does not add maintenance burden on said library.
> 
> I don't think anything blocks you from achieving your way you just have to stick to the form parameters and values the way it is.

 I understand that you are concerned about adding maintenance burden to the library with the suggested change.

I want to assure you that the change I proposed is a simple configuration option that does not require any changes to existing code or functionality. It simply allows the developer to specify whether the 'confirm' parameter should accept only boolean values, and to specify the name of the confirmation parameter if desired. 

This adds minimal complexity to the library and should not require any additional maintenance effort.
Additionally, The change is well-tested and does not cause any issues.😉

I appreciate your consideration of this suggestion, and I hope that it will be helpful to other developers who are using the library as an **API** solution, as it would allow us to use the device flow feature in a more flexible way. 

Thank you!

PR #1212, opened 2022-12-21
Allow engines greater or equal to 18

jonschlinkert:
This pull request updates the `engine` property in package.json to `^18`, to allow versions equal to or greater than 18. Which fixes the following:

image

FWIW, this error message is displayed more than any other message in my terminal. It's quite frustrating debugging with this message taking up most of the terminal.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,6 @@
     "timekeeper": "^2.2.0"
   },
   "engines": {
-    "node": "18"
+    "node": "^18"
   }
 }


panva:
The engines entry has nothing to do with that text in your terminal. 

Just use an actual 18.x LTS release and the message will go away.

jonschlinkert:
> The engines entry has nothing to do with that text in your terminal.

So you're hardcoding that message to a specific package.json version in your code, instead of using a variable?  Good to know.

jonschlinkert:
To be clear, you're saying that only 18.x LTS versions are supported, but not later versions. So I need to downgrade to use this without the warnings?

panva:
Yes. It has been the case for a very long time that I only support LTS releases in oidc-provider.

PR #1213, opened 2022-12-24
example: add prisma adapter

Mostafatalaat770:
Add a fully typed Prisma adapter with a reference to an example with the oidc-provider consuming it.

### DIFF
--- a/example/adapters/contributed/prisma.ts
+++ b/example/adapters/contributed/prisma.ts
@@ -1,0 +1,173 @@
+/*
+ * A simple Fully typed adapter for oidc-provider that uses Prisma, you can
+ * find more information about the adapter with an expressjs example here:
+ * https://github.com/Mostafatalaat770/node-oidc-provider-prisma-adapter
+ *
+ *
+ * for more information, you can contact me on:
+ * - GitHub: https://github.com/Mostafatalaat770
+ * - LinkedIn: https://www.linkedin.com/in/mostafatalaat770/
+ * - Email: mostafatalaat770@gmail.com
+ *
+ *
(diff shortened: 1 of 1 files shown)


pi0neerpat:
Thank you! I think the example repo is still private https://github.com/Mostafatalaat770/node-oidc-provider-prisma-adapter

Mostafatalaat770:
> Thank you! I think the example repo is still private https://github.com/Mostafatalaat770/node-oidc-provider-prisma-adapter

Thanks for pointing out that! I totally forgot to make it public

amydevs:
thank you for this! was about to write my own and thought to search through the pull requests

Mostafatalaat770:
> thank you for this! was about to write my own and thought to search through the pull requests

you're welcome! glad I could help.

pi0neerpat:
I used your adapter to create one here https://github.com/UseKeyp/oauth2-server-redwood/blob/d9c00bfdbbe1c18e684aa09343ee27a6e26701bc/api/db/schema.prisma#L53. I have it working now in my Redwood app. I'm currently in the process of adding dynamic client generation.

If you're free, I'd love to hear your input on why you made this schema decisions

```graphql
model Oidc {
   id         String    @id
   type       Int 
   @@unique([id, type])
   @@map("oidc_model")
}

# Instead of 

model Oidc {
   id        String    @default(cuid())
   type       Int
   @@id([id, type])
}
```

Mostafatalaat770:
@pi0neerpat thanks for your feedback! you are right, your suggestion makes more sense, and both will work as intended, I didn't have anything in mind when deciding the schema, but yea I would change it in the future to have the composite key.

PR #1219, opened 2023-04-10
Error forward

PauloFerreira25:
Hello, please review my pull request.

What do you identify as a problem:
When the request is not made via HTML, it is impossible to capture the error event in order to use a Koa Error Handling, as the error is converted to a message in the body.

The ideal would be to send the error forward and thus be recaptured.

### DIFF
--- a/lib/helpers/initialize_app.js
+++ b/lib/helpers/initialize_app.js
@@ -218,11 +218,5 @@ export default function initializeApp() {
 
   this.app.use(proxyWarning);
   this.app.use(error(this));
-  this.app.use(async (ctx, next) => {
-    await next();
-    if (ctx.status === 404 && ctx.message === 'Not Found') {
-      throw new InvalidRequest(`unrecognized route or not allowed method (${ctx.method} on ${ctx.path})`, 404);
-    }
-  });
   this.app.use(router.routes());
 }
--- a/lib/shared/error_handler.js
+++ b/lib/shared/error_handler.js
@@ -27,6 +27,7 @@ export default function getErrorHandler(provider, eventName) {
     } catch (err) {
       const out = errOut(err);
       ctx.status = err.statusCode || 500;
+      let doThrow = false
 
       if (err.expose && !(err instanceof ReRenderError)) {
         debug('path=%s method=%s error=%o detail=%s', ctx.path, ctx.method, out, err.error_detail);
@@ -51,13 +52,16 @@ export default function getErrorHandler(provider, eventName) {
         const renderError = instance(provider).configuration('renderError');
         await renderError(ctx, out, err);
       } else {
-        ctx.body = out;
+        doThrow = true
       }
 
       if (out.error === 'server_error') {
         provider.emit('server_error', ctx, err);
       } else if (eventName) {
         provider.emit(eventName, ctx, err);
+      }
+      if(doThrow){
+        throw err;
       }
     }
   };


panva (changes requested):
Hello @PauloFerreira25 

I've reviewed your PR. It fails to clearly explain the intended goal and the problem statement. It also lacks updates or new additions to the test suite to reflect new behaviour.

PauloFerreira25:
Hi @panva, I'll try to express myself better.

The problem:
In the current code, there are 2 treatments to make the error response more "human" for the end user in cases of errors.

Being the first:
https://github.com/panva/node-oidc-provider/blob/main/lib/helpers/initialize_app.js#L223
This captures anything that is 404 (Not found) and turns it into an InvalidRequest.
What's the problem with this:
If a route is created in the same koa that the oidc-provider is being used, for example, a /user/:id route and the return returns 404 - Not Found, because the id does not exist. It will be turned into InvalidRequest.

Second:
Any route that returns an error ends up arriving in the code:
https://github.com/panva/node-oidc-provider/blob/main/lib/shared/error_handler.js#L20

What is the problem:
When a call is made outside a browser (such as curl), the response header format goes like anything other than HTML.
Arriving at this else:
https://github.com/panva/node-oidc-provider/blob/main/lib/shared/error_handler.js#L54

This else transforms a type error into a type string (ctx.body), and does not resubmit this error. That is, it "finalizes" the error.

Doing this makes it impossible to use the model that Koa describes the "default error handler":
https://github.com/koajs/koa/blob/master/docs/error-handling.md#default-error-handler

Because an error is never sent to be caught.

I tried the idea of looking at the ctx.body to find out if it's an error or not, but it's a very high CPU cost to keep parsing the string in type again.


About the tests:
I need to study more how to produce them. If you have any material to help me with this I would be grateful.


Please forgive me for the delay in responding.

Thank you very much for your attention.

panva:
> If a route is created in the same koa that the oidc-provider is being used, for example, a /user/:id route and the return returns 404 - Not Found, because the id does not exist. It will be turned into InvalidRequest.

This is not what i'm observing, https://github.com/panva/node-oidc-provider/blob/v8.x/example/standalone.js shows how e.g. interactions are added to the provier's koa app dispatcher and they do get used.

> Doing this makes it impossible to use the model that Koa describes the "default error handler":

I believe you shouldn't be doing anything default error handler on the provider koa instance. Do it in your own and let the provider do its thing, hook into its error events if you need to observe errors.

PauloFerreira25:
> This is not what i'm observing, https://github.com/panva/node-oidc-provider/blob/v8.x/example/standalone.js shows how e.g. interactions are added to the provier's koa app dispatcher and they do get used.

About the 404 being converted to InvalidRequest.
This test shows exactly the problem point I'm expressing:
https://github.com/panva/node-oidc-provider/blob/main/test/routing/routing.test.js#L17

Literally gets a 404 waiting if application/json:
https://github.com/panva/node-oidc-provider/blob/main/lib/helpers/initialize_app.js#L224

> I believe you shouldn't be doing anything default error handler on the provider koa instance. Do it in your own and let the provider do its thing, hook into its error events if you need to observe errors.

I'm not wanting to intervene in the provider's error handling using the 'default-error-handler'.
I'm wanting to capture the errors of routes that are in the same koa as the provider.
Because today, any ctx.thow is converted to 200-OK with the unformatted error in the body.

panva:
I'm not seeing it. Same result regardless if I run example/koa or example/standalone.

```diff
diff --git a/example/routes/koa.js b/example/routes/koa.js
index 0c326611..a8d845bd 100644
--- a/example/routes/koa.js
+++ b/example/routes/koa.js
@@ -42,6 +42,10 @@ export default (provider) => {
     }
   });
 
+  router.get('/anything', (ctx) => {
+    ctx.throw(404, 'id not found.');
+  });
+
   router.get('/interaction/:uid', async (ctx, next) => {
     const {
       uid, prompt, params, session,
```

```
curl -iX GET https://op.panva.me/anything
HTTP/2 404 
content-type: text/plain; charset=utf-8
date: Thu, 13 Apr 2023 15:13:15 GMT
server: Caddy
content-length: 13

id not found.
```

panva:
Feel free to invest more time into this, I'm not sure I'm able to help. I'm going to go ahead and close this PR, as you can see this particular change breaks conformance of almost every error response. If you figure it out, add tests and not break existing ones (which aren't related to just this catch-all), feel free to open another PR.

PauloFerreira25:
Yes, I noticed later the errors in the tests. (My apologies)

Looking more fondly. I believe that my approach to PR can be improved. I believe that there may be an error hook to allow us not to fall into this scenario and leave it free for those who are going to use it to create their own custom errors.

Currently, the solution I adopted was to segregate 2 koa middleware in the same project. Thus, other APIs do not compete with the oidc-provider flow.

Again, thank you very much for your attention.

PR #1221, opened 2023-05-06
fix: ignore post_logout_redirect_uris when logout is disabled

manvydasu:
Hello, I think there is a slight problem wtih `rpInitiatedLogout` configuration.

inside the `client_schema.js` there is this check at line 79

```
  if (features.rpInitiatedLogout.enabled) {
    RECOGNIZED_METADATA.push('post_logout_redirect_uris');
  }
```
However, it looks like it does nothing, because `post_logout_redirect_uris` is already included in the default `RECOGNIZED_METADATA` array.  In order for flag `features.rpInitiatedLogout.enabled` to properly work, default recognized metadata properties should be changed.

I can't run tests locally due to `structuredClone` error, so haven't verified yet if tests don't break.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -439,6 +439,7 @@ location / {
   - [requestObjects](#featuresrequestobjects)
   - [resourceIndicators ❗](#featuresresourceindicators)
   - [revocation](#featuresrevocation)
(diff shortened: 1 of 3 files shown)


panva:
Nice catch, thank you @manvydasu

PR #1225, opened 2023-07-03
Router prefix

lbennett-stacki:
> fix: set core koa router prefix to correct routes when mounted

I'm mounting a provider in a nextjs app under the path `/api/oidc`. I noticed the discovery paths were wrong, but also some generated paths like ones used by `interactionFinished` would be incorrect. Making this change put everything in line.

Is this something that would be accepted with tests?

prefix option docs: https://github.com/ZijianHe/koa-router#new-routeropts

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -290,8 +290,10 @@ provider.use(async (ctx, next) => {
 The following snippets show how a provider instance can be mounted to existing applications with a
 path prefix `/oidc`.
 
-Note: if you mount oidc-provider to a path it's likely you will have to also update the 
-[`interactions.url`](#interactionsurl) configuration to reflect the new path.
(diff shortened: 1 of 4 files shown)


panva:
Please make sure to add tests, e. g. similar to the existing ones that your changes broke. Obviously your changes also must not break existing ones.

panva:
Also,

> I'm mounting a provider in a nextjs app under the path /api/oidc.

how are you doing so? Please provide a sample minimal repo with this setup.

lbennett-stacki:
I'll have a look at finishing tests on the weekend. Test suite looks great. Pushed an example for next mount but updates to come.

panva:
This has became stale, closing for now.

PR #1228, opened 2023-08-17
docs: Fix typo in `result.login` specs

SPodjasek:
Just a tiny fix for typo


SPodjasek:
And a question in regard of this PR, I've noticed that after saving my editor strips many trailing white-spaces in `README.md` (as specified in `.editorconfig`) - but I've forced it not to do so. Should I commit with all this white-spaces removed?

panva:
Thank you for spotting and fixing this ;)

PR #1233, opened 2023-11-23
feat: add code to be compatible with Nest configuration

locene:
https://docs.nestjs.com/faq/raw-body


panva:
This would belong to the block below that warns, I believe it is a good candidate to extract into its own helper function.

panva:
The use of "raw body" that you linked still falls under a warning, since the request stream has already been consumed when it would get accessed. To get away from the warning, have nest not consume the request stream, or live with a one time warning, it doesn't seem like this is a blocker and so probably not worth the effort to refactor.

PR #1234, opened 2023-11-23
feat: add code to be compatible with Nest configuration

locene:
https://docs.nestjs.com/faq/raw-body


panva:
I will consider this for the main branch but will not be doing backports.

locene:
Starting from v8.0.0, oidc-provider is an ESM-only module. That's why I submitted the PR to provide an alternative for those with specific requirements (including some of my projects). Could you please reconsider?

PR #1237, opened 2023-12-01
fix: reject client JSON Web Key Set `null` value

KalleV:
Prevents "jwks" set to null in the [client metadata](https://github.com/panva/node-oidc-provider/tree/main/docs#clients) instead of a TypeError such as:
```
TypeError: Cannot read properties of null (reading 'keys')\n    at validateJWKS
(node_modules/.pnpm/oidc-provider@8.4.1/node_modules/oidc-provider/lib/models/client.js:41:29
```

### DIFF
--- a/lib/models/client.js
+++ b/lib/models/client.js
@@ -37,7 +37,7 @@ const fingerprint = (properties) => hash(properties, {
 });
 
 const validateJWKS = (jwks) => {
-  if (jwks !== undefined) {
+  if (jwks !== undefined && jwks !== null) {
     if (!Array.isArray(jwks.keys) || !jwks.keys.every(isPlainObject)) {
       throw new InvalidClientMetadata('client JSON Web Key Set is invalid');
     }
--- a/lib/models/client.js
+++ b/lib/models/client.js
@@ -37,8 +37,8 @@ const fingerprint = (properties) => hash(properties, {
 });
 
 const validateJWKS = (jwks) => {
-  if (jwks !== undefined && jwks !== null) {
-    if (!Array.isArray(jwks.keys) || !jwks.keys.every(isPlainObject)) {
+  if (jwks !== undefined) {
+    if (!Array.isArray(jwks?.keys) || !jwks.keys.every(isPlainObject)) {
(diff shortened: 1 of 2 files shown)


lib/models/client.js
```diff
@@ -37,7 +37,7 @@ const fingerprint = (properties) => hash(properties, {
 });
 
 const validateJWKS = (jwks) => {
-  if (jwks !== undefined) {
+  if (jwks !== undefined && jwks !== null) {
```
panva, line 40:
```suggestion
  if (!isPlainObject(jwks)) {
```

panva (changes requested):
I

lib/models/client.js
```diff
@@ -37,7 +37,7 @@ const fingerprint = (properties) => hash(properties, {
...
     if (!Array.isArray(jwks.keys) || !jwks.keys.every(isPlainObject)) {
```
panva, line 41:
```suggestion
  if (jwks !== undefined) {
    if (!Array.isArray(jwks?.keys) || !jwks.keys.every(isPlainObject)) {
```

panva (changes requested):
null is not a valid JSON value for the `jwks` property, so it should be properly rejected instead.

You should also add a test for this

https://github.com/panva/node-oidc-provider/blob/a0af4c45343c598932676fadd78625b4d9a42a10/test/configuration/client_metadata.test.js#L1753

```js
rejects(this.title, 1, 'client JSON Web Key Set is invalid');
```

KalleV:
> null is not a valid JSON value for the `jwks` property, so it should be properly rejected instead.
> 
> You should also add a test for this in
> 
> https://github.com/panva/node-oidc-provider/blob/a0af4c45343c598932676fadd78625b4d9a42a10/test/configuration/client_metadata.test.js#L1753
> 
> Something along the lines of
> 
> ```js
> rejects(this.title, null, 'client JSON Web Key Set is invalid');
> ```

Thanks! I'll add that test.

panva (approved):

panva:
Thank you @KalleV

PR #1238, opened 2023-12-03
test(client_metadata): add empty string and null checks to client_name test

KalleV:
Add assertions to test the empty string and null value checks for the "client_name" metadata. For example:
```
Error: InvalidClientMetadata received invalid_client_metadata client_name must be a non-empty string if provided
```


KalleV:
> IIRC these assertions are already part of `mustBeString`

Ah, okay. I see that now. Thanks, I'll close this one.

PR #1244, opened 2023-12-27
[7.x] Fix cve-2022-25883 for non ESM-only projects

AntoineDewaele:
Hello,

I would like to fix this CVE for 7.x versions : https://avd.aquasec.com/nvd/2022/cve-2022-25883/

I use this package on a Nest project and there is still no plan for Nest to add ESM-only packages support (https://github.com/nestjs/nest/pull/8736), so it's impossible for Nest project to migrate to v8.4.3 that fix this CVE

I check on this commit https://github.com/panva/node-oidc-provider/commit/9023e23c0c567ab712655595a7e4fb06c3907def to replicate changes

It's my first contribution on this package, so sorry if i did not respect standard, process or whatever.

Thanks a lot 🙏

### DIFF
--- a/lib/shared/cors.js
+++ b/lib/shared/cors.js
@@ -20,7 +20,13 @@ function checkClientCORS(ctx, client) {
 }
 
 module.exports = ({ clientBased = false, ...options }) => {
-  const builtin = cors({ keepHeadersOnError: false, ...options });
+  const builtin = cors({
+    keepHeadersOnError: false,
+    origin(ctx) {
+      return ctx.get('Origin') || '*';
+    },
+    ...options,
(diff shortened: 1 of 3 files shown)


panva:
Hello @AntoineDewaele 

You can just ignore the CVE.

The way oidc-provider uses CORS is per-endpoint and where appropriate enabled per your own policy. The discovery and jwks endpoints are intentionally CORS-open, the rest is preflight-enabled and actual request depends on your own policy in the `clientBasedCORS` helper. The behaviour remains the same before and after the patch.

I am not doing backports to v7.x unless there's an actual vulnerability.

PR #1245, opened 2024-01-17
fix: missing opening tag in response html

charsleysa:


panva:
@charsleysa thank you, nice catch!

panva:
Sorry for the noise, this is now merged to main. Thank you

PR #1257, opened 2024-04-17
allow to override discovery urls, to redirect auth endpoint to a fron…

wzrdtales:
This allows to override discovery urls, to redirect auth endpoint to a frontend. The reason for this is that we're laying an independent frontend on top, from there we're passing through the necessary params to the backend. For this to work, the well-known config must return a different path for the auth endpoint in specific.

### DIFF
--- a/lib/actions/discovery.js
+++ b/lib/actions/discovery.js
@@ -5,11 +5,11 @@ import instance from '../helpers/weak_cache.js';
 
 export default function discovery(ctx, next) {
   const config = instance(ctx.oidc.provider).configuration();
-  const { features } = config;
+  const { features, discovery: { urls: discoveryUrls } } = config;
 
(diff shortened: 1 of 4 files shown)


panva:
Hi @wzrdtales 

You can use a [post-middleware](https://github.com/panva/node-oidc-provider/blob/v8.x/docs/README.md#pre--and-post-middlewares) to replace the entire response of the discovery document.

panva:
Sorry I won't be accepting proprietary changes to protocol-defined behaviours.

wzrdtales:
How is this proprietary? `x-requested-with` is a very well-defined open behavior and this is to make this module able to be combined with a frontend that is **NOT** tied to the internal renderer! That has nothing to do with proprietary... How do you even get to this conclusion?

PR #1258, opened 2024-04-23
fix: ensure each individual resource indicator is a valid URI

charIeszhao:
Fix a bug in `checkResource` util function.

When checking if the resource identifier is a valid URL, we should iterate the resource array with a for-loop, and check if each identifier in the `resource` array is valid, by utilizing the `new URL()` constructor.

However, the origin code seems to mistakenly pass the entire `resource` array into the URL constructor, resulting unexpected behavior.

Fixes: https://github.com/logto-io/logto/issues/5779

### DIFF
--- a/lib/shared/check_resource.js
+++ b/lib/shared/check_resource.js
@@ -55,7 +55,7 @@ export default async function checkResource(ctx, next) {
   for (const identifier of resource) {
     let href;
     try {
-      ({ href } = new URL(resource));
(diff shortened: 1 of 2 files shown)


panva:
You're absolutely right. This only works when there's a single element due to how new URL calling toString() on non-string inputs.

Let me just add a regression test.

charIeszhao:
Thanks! Glad I can help.

PR #1260, opened 2024-04-26
Keep capacitor uri changes but remove prettier formats

StefanZivkovic:


StefanZivkovic:
made by mistake, ignore, or delete PR if you can

PR #1283, opened 2024-11-13
docs: fix `hapi` mounting

wxh06:
Current behavior: `[/].well-known/openid-configuration` -> `[/oidc].well-known/openid-configuration`


panva (dismissed):

panva:
@wxh06 the current example is being tested in CI, your fix doesn't break CI, but I would be keen to know more about what this is fixing... From what i'm looking at there's no issue... 🤷

PR #1287, opened 2025-01-16
fix: allow specifying the user-agent header for outgoing requests

pvieira-diligent:
Specifying the `user-agent` header for outgoing requests was introduced in https://github.com/panva/node-oidc-provider/commit/95f24ef6365ae54375f7459cec19fb1c6b8ba291.

This PR is fixing an issue with that commit: the `user-agent` header value set in the `httpOptions` config is not really being used, and the http agent defaults to not sending the `user-agent` HTTP header.

```javascript
  const helperOptions = pickBy({ signal, agent, dnsLookup }, Boolean);
  // ...
  options.headers['user-agent'] = helperOptions['user-agent']; // at this point, helperOptions['user-agent'] is always undefined
```

In practice, due to the missing `user-agent`, WAFs could block requests to the `jwks_uri` endpoint during a client authentication.

### DIFF
--- a/lib/helpers/request.js
+++ b/lib/helpers/request.js
@@ -19,6 +19,7 @@ export default async function request(options) {
     signal = AbortSignal.timeout(2500),
     agent = options.url.protocol === 'http:' ? http.globalAgent : https.globalAgent,
     dnsLookup = dns.lookup,
+    'user-agent': userAgent = undefined,
   } = instance(this).configuration('httpOptions')(new URL(options.url));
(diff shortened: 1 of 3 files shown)


panva:
Clearly i should've added a test then, apologies. Can you add a test now?

pvieira-diligent:
> Clearly i should've added a test then, apologies. Can you add a test now?

I've added a couple of tests already. 

Are you expecting different ones? I'm happy to add more if I'm missing any scenarios.

panva:
Thank you @pvieira-diligent

PR #1293, opened 2025-02-19
Update err_out.js

yashzawar02:
Added error_detail property in error response, if the response type is json.

This property was not getting propagated to client in response.


panva:
error_detail is not intended to go over the protocol messages, it's intended to be something available to the operator and through out of scope means to the client if the operator makes it so, e.g. through client specific logs available to administrators.

PR #1328, opened 2025-04-24
Remove toString() function type check

grahhnt:
In alternative runtimes Function#toString() doesn't equal `function Function() { [native code] }` exactly, which causes checkTTL() to fail

This can be seen using Bun 1.2.10 and oidc-provider@9.0.1

### DIFF
--- a/lib/helpers/configuration.js
+++ b/lib/helpers/configuration.js
@@ -370,9 +370,7 @@ class Configuration {
       let valid = false;
       switch (typeof value) {
         case 'function':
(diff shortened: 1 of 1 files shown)


panva:
This module is not intended to be used outside of Node.js. It is described as

> OpenID Certified™ OAuth 2.0 Authorization Server implementation for Node.js

It is not being tested in other runtimes. While I will not go out of my way to block usage in non-Node.js runtimes, I will also not be making "fixes" towards the goal of being able to.

PR #1357, opened 2025-09-18
Allow auth code flow for clients with implicit response types

flipscholtz:
When a client registers with a response type requiring an implicit flow, they should still be able to ask for an authorization_code grant type if they want.
i.e. clients can say: "I accept responses one of which can only be obtained via the implicit flow, but I don't want to use the implicit flow, I only want to use auth code flow"
This is needed by LibreChat, for example.

This PR adds a new config flag (`relaxImplicitGrantRequirement`) that allows the above scenario by relaxing oidc-provider's default validation rules.


flipscholtz:
Closing, accidentally created against upstream.

